PR #3, opened 2016-10-12
Don't crash on errors

atifaziz:
`dotnet script` _crashes_ on any error (like bad command-line arguments).

For example on Windows, assuming a script `foo.csx` in the current directory, running the following:

    dotnet script foo.csx -c

will end with a crash:

image

and the following message on the console:

```
Specify --help for a list of available options and commands.
Unhandled Exception: Microsoft.Extensions.CommandLineUtils.CommandParsingException: Missing value for option 'configuration'
   at Microsoft.Extensions.CommandLineUtils.CommandLineApplication.Execute(String[] args)
   at Dotnet.Script.Program.Main(String[] args)
```

This PR fixes the issue by handling all exceptions at program entry-point level so it can terminate normally but signal an error condition to any caller with a non-zero exit code (as per convention for OS processes).

### DIFF
--- a/src/Dotnet.Script/Program.cs
+++ b/src/Dotnet.Script/Program.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using System.Collections;
 using System.Collections.Generic;
 using System.Diagnostics;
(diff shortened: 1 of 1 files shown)


filipw:
thanks a lot

filipw:
https://www.nuget.org/packages/Dotnet.Script/0.2.0-beta

PR #6, opened 2016-10-12
restore some sanity

filipw:
no functional changes, just get the crap out of `Program.cs` and lay ground for a future development in a more structured way

### DIFF
--- a/src/Dotnet.Script/ScriptRunner.cs
+++ b/src/Dotnet.Script/ScriptRunner.cs
@@ -1,0 +1,144 @@
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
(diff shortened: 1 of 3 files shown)


src/Dotnet.Script/ScriptRunner.cs
```diff
@@ -0,0 +1,144 @@
...
+
+        private static readonly IEnumerable<string> DefaultNamespaces = new[]
+        {
+            "System",
+            "System.IO",
+            "System.Linq",
+            "System.Collections.Generic"
+        };
+
+        public async Task Execute<TReturn>(ScriptContext context)
```
atifaziz: `TReturn` is never used

PR #9, opened 2016-10-12
return script value from the runner

filipw:
for now we throw it away, but it will be helpful trying to use the runner with other use cases (i.e. web app)


filipw:
rebased

PR #10, opened 2016-10-12
Use STDERR for debug & runner output

atifaziz:
Debug and runner output should use `STDERR` only so that it doesn't interfere with _real_ output from script being run. For example, if the script produces a CSV, the runner will currently produced corrupted output (especially with `--debug`). This PR fixes this by exclusively using `STDERR` for the runner.

### DIFF
--- a/src/Dotnet.Script/Program.cs
+++ b/src/Dotnet.Script/Program.cs
@@ -53,7 +53,7 @@ namespace Dotnet.Script
 
         private static void RunScript(string file, string config, bool debugMode, List<string> scriptArgs)
         {
-            var runner = new ScriptRunner();
+            var runner = new ScriptRunner(Console.Error);
             var context = new ScriptContext(file, config, debugMode, scriptArgs);
 
             runner.Execute<object>(context).GetAwaiter().GetResult();
--- a/src/Dotnet.Script/ScriptRunner.cs
+++ b/src/Dotnet.Script/ScriptRunner.cs
@@ -30,19 +30,30 @@ namespace Dotnet.Script
             "System.Collections.Generic"
         };
 
+        readonly TextWriter _stderr;
+
+        public ScriptRunner(TextWriter stderr)
+        {
+            _stderr = stderr ?? TextWriter.Null;
+        }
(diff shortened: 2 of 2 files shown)


filipw:
yeah this is a good idea. I was thinking of adding some sort of an abstraction to redirect output and give option to process it and so on. 
That abstraction could then be also passed into the script via `ScriptingHost` allowing the script to write to a place that the host defines. 

but that's long term, for now `TextWriter` is more than enough.

atifaziz:
> That abstraction could then be also passed into the script via ScriptingHost allowing the script to write to a place that the host defines.

I would generally discourage taking a dependency on [`ScriptingHost`][scrhost] so that a `.csx` can remain runner-independent by as much as possible. For simple stuff like [`ScriptingHost.ScriptDirectory`][scrdir] and [`ScriptingHost.ScriptPath`][scrpath], using environment variables would be best. For advances cases (which may never happen in practice, mind you), [`ScriptingHost.ScriptAssembly`][scrasm] is just fine. For output, you can just redirect the standard output to wherever the host wants.

[scrhost]: https://github.com/filipw/dotnet-script/blob/fe80a9eefd088f190dcbfc311454979542f90d92/src/Dotnet.Script/ScriptingHost.cs
[scrdir]: https://github.com/filipw/dotnet-script/blob/fe80a9eefd088f190dcbfc311454979542f90d92/src/Dotnet.Script/ScriptingHost.cs#L10
[scrpath]: https://github.com/filipw/dotnet-script/blob/fe80a9eefd088f190dcbfc311454979542f90d92/src/Dotnet.Script/ScriptingHost.cs#L12
[scrasm]: https://github.com/filipw/dotnet-script/blob/fe80a9eefd088f190dcbfc311454979542f90d92/src/Dotnet.Script/ScriptingHost.cs#L14

filipw:
Yeah agree with you.

That said, it is entirely opt in. Where it's helpful is when dealing with things are impossible to achieve in scripting without the active help of a host - for example running an ASP.NET Core HTTP service from a script.

PR #12, opened 2016-10-12
Throw & fail if script has compilation errors

atifaziz:
A script that fails to compile should not be considered successful, yet it is. Suppose the following script named `foo.csx`:

```c#
C0nsole.WriteLine("Hello from foo.csx");
```

The following command line on the shell should not print OK, yet it does:

```
$ dotnet script foo.csx && echo OK
There is an error in the script.(1,1): error CS0103: The name 'C0nsole' does not exist in the current context
OK
```

This PR will fix the issue by first printing all diagnostics to `STDERR` and then throwing an exception:

```
$ dotnet script foo.csx && echo OK
(1,1): error CS0103: The name 'C0nsole' does not exist in the current context
Script compilation failed due to one or more errors.
```

Note that `echo OK` does not run since `dotnet script` ends with a non-zero exit code now.


filipw:
thanks!

PR #13, opened 2016-10-12
Added support for debugging

filipw:


filipw:
We can now debug scripts using Visual Studio Code.

filipw:
I might have broken something else in the process 😎

PR #15, opened 2016-10-14
Convert late-bound calls into strong-typed invocations

atifaziz:
As I was going over the code, I took the opportunity to turn the late-bound calls (into Roslyn internals) into strong-typed ones by creating `Func<…>` delegates over the reflected method. It makes the signature clearer. For example:

```c#
var loaderAseemblyLoadMethod = compilationContext.Loader.GetType().GetMethod("LoadAssemblyFromStream", BindingFlags.Instance | BindingFlags.NonPublic);
var assembly = loaderAseemblyLoadMethod.Invoke(compilationContext.Loader, new[] { peStream, pdbStream }) as Assembly;
```

becomes:

```c#
var assembly = compilationContext.Host.ScriptAssembly =
    // https://github.com/dotnet/roslyn/blob/version-2.0.0-beta4/src/Scripting/Core/Hosting/AssemblyLoader/InteractiveAssemblyLoader.cs#L111
    compilationContext.Loader.Invoke<Stream, Stream, Assembly>(
        "LoadAssemblyFromStream", BindingFlags.NonPublic,
        peStream, pdbStream);
```

I also dropped links into Roslyn sources for each internal as a reference.

The dynamic bindings in `ScriptStateExtensions.GetScriptAssembly` have been replaced with use of stronger (albeit non-generic) types.

### DIFF
--- a/src/Dotnet.Script/DebugScriptRunner.cs
+++ b/src/Dotnet.Script/DebugScriptRunner.cs
@@ -8,7 +8,6 @@ using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.CodeAnalysis;
 using Microsoft.CodeAnalysis.Emit;
-using Microsoft.CodeAnalysis.Scripting;
 
 namespace Dotnet.Script
 {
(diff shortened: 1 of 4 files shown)


filipw:
thanks!

src/Dotnet.Script/DebugScriptRunner.cs
```diff
@@ -34,9 +33,11 @@ public override Task<TReturn> Execute<TReturn>(ScriptContext context)
 
             foreach (var syntaxTree in compilationContext.Compilation.SyntaxTrees)
             {
+                // https://github.com/dotnet/roslyn/blob/version-2.0.0-beta4/src/Compilers/CSharp/Portable/Syntax/CSharpSyntaxTree.ParsedSyntaxTree.cs#L19
```
adamralph: The problem with all these URLs is that they are not absolute. I.e. what is in line 19 of that file will change over time. Perhaps the file itself will move.

You should link to the specific version at the time of writing, e.g. https://github.com/dotnet/roslyn/blob/eb964d42c8f8e763e4379f1e33c3cf5547692a3f/src/Compilers/CSharp/Portable/Syntax/CSharpSyntaxTree.ParsedSyntaxTree.cs#L19

filipw: good point yes

filipw: actually - no, these links point to a tag, so they are OK

adamralph: True, I didn't spot that. They will be OK as long as the tag doesn't move. 😉

adamralph: Yes, it should be fine.

PR #16, opened 2016-10-14
make ScriptingHost inherit from InteractiveScriptGlobals

adamralph:
This way, *any* script written for csi.exe will run in dotnet-script.

### DIFF
--- a/src/Dotnet.Script/ScriptRunner.cs
+++ b/src/Dotnet.Script/ScriptRunner.cs
@@ -8,6 +8,7 @@ using System.Text;
 using System.Threading.Tasks;
 using Microsoft.CodeAnalysis;
 using Microsoft.CodeAnalysis.CSharp.Scripting;
+using Microsoft.CodeAnalysis.CSharp.Scripting.Hosting;
 using Microsoft.CodeAnalysis.Scripting;
 using Microsoft.CodeAnalysis.Scripting.Hosting;
 using Microsoft.CodeAnalysis.Text;
@@ -118,12 +119,16 @@ namespace Dotnet.Script
                                                     diagnostics);
             }
 
-            var host = new ScriptingHost
+            var host = new ScriptingHost(Console.Out, CSharpObjectFormatter.Instance)
             {
                 ScriptDirectory = directory,
                 ScriptPath = context.FilePath,
-                Args = context.ScriptArgs,
             };
+
+            foreach (var arg in context.ScriptArgs)
+            {
+                host.Args.Add(arg);
+            }
 
             return new ScriptCompilationContext<TReturn>(compilation, script, host, sourceText, loader);
         }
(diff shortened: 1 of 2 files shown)


adamralph:
You may want to consider removing `ScriptingHost` entirely, so that scripts are portable in other direction, i.e. any script written for dotnet-script will run in csi.exe

Personally I believe all CSX runners should use a standard set of globals so that scripts are completely portable. The closest thing we have to such a standard is `InteractiveScriptGlobals`.

filipw:
yeah makes sense for me, although it is an opt in anyway?

Personally I don't care about anything on it - except for the current assembly - which is really needed if you want to do more advanced things with the script

adamralph:
I think it may be best to send a PR to Roslyn to get that (and anything else which is useful) added to `InteractiveScriptGlobals`.

atifaziz:
> Personally I don't care about anything on it - except for the current assembly

@filipw Surprised you didn't pick `Args` because I'd expect to win the popularity context by far from anything else on the host globals; unless, of course one is hard-coding everything into the script because it's just that, _a script_. However, another big motivation for C# scripts is often to avoid the whole program/project/compilation ceremony for simple automation cases & command-line args help to build some flexibility and options (aka configuration). Ideally one could just use [`Environment.GetCommandLineArgs`](https://msdn.microsoft.com/en-us/library/system.environment.getcommandlineargs.aspx) but trouble there is that you have to assume and parse out some host args (e.g. by convention of everything after first `--`) to get to your real args & that's just plain ugly to bake into every script.

> I think it may be best to send a PR to Roslyn to get that (and anything else which is useful) added to `InteractiveScriptGlobals`.

And perhaps have globals factored into interfaces for standardization, discovery and so on. In fact, could have that as a separate package run by the community and ask Roslyn to take it as a dependency & implement on `InteractiveScriptGlobals`. The idea would be then be to enable writing code like this:

```c#
var args = (this as IScriptArgs)?.Args ?? new string[0];
```

or the following if you want to require the host feature & crash & burn if absent:

```c#
var args = ((IScriptArgs) this).Args;
```

filipw:
> Personally I don't care about anything on it - except for the current assembly

I meant except for the things already available on `InteractiveScriptGlobals` which already contains args.

adamralph:
@filipw `System.Reflection.MethodInfo.GetCurrentMethod().DeclaringType.Assembly` seems to work for getting the script assembly.

image

atifaziz:
@adamralph I think you mean [`MethodBase.GetCurrentMethod`](https://msdn.microsoft.com/en-us/library/system.reflection.methodbase.getcurrentmethod.aspx?). Either way, this is definitely better than [relying on Roslyn internals as the current implementation does](https://github.com/filipw/dotnet-script/blob/194ce42b357c0e6dbbc7647792228652c691f248/src/Dotnet.Script/ScriptStateExtensions.cs#L12).

adamralph:
@atifaziz you're correct, I should have used `MethodBase`.

I've raised a PR to remove the redundant member - https://github.com/filipw/dotnet-script/pull/24

PR #18, opened 2016-10-14
Fix to not show help at the end of script run

atifaziz:
Help should display only when zero/no args were supplied but right now it's displaying unconditionally. This PR fixes ba0e4bad0f34844b23059492bc0847b2fbdc3f0b that bork this.


filipw:
ugh - thanks

PR #17, opened 2016-10-14
Added support for code execution + bug fixes

filipw:
- added support for code execution (fileless) i.e. `dotnet script eval Console.WriteLine(\"Hi\");`. This is based on `node -e console.log(\"foo\")`
 - show help if running without any args i.e. `dotnet script`
 - fixed a bug introduced by @adamralph 😄 `c:\script-test2\foo.csx(4,5): error CS0012: The type 'InteractiveScriptGlobals' is defined in an assembly that is not referenced. You must add a reference to assembly 'Microsoft.CodeAnalysis.Scripting, Version=2.0.0.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35'.`
 - removed everything from `ScriptingHost` except for `ScriptAssembly`

PR #20, opened 2016-10-14
Fix potential file leak & locking

atifaziz:
The opened `FileStream` is not closed/disposed anywhere in a deterministic fashion and in face of unclear ownership for now, it seems safest to just buffer the content.


atifaziz:
Never mind. Looks like this issue went away with #17 and I was still on the master before #17 was merged.

Issue #11, opened 2016-10-12
More intuitive command-line usage


atifaziz:
Current usage:

```
Usage:  [arguments] [options]

Arguments:
  script  Path to CSX script

Options:
  -a |--arg <args>                     Arguments to pass to a script. Multiple values supported
  -c |--configuration <configuration>  Configuration to use. Defaults to 'Release'
  -d | --debug                         Enables debug output.
  -? | -h | --help                     Show help information
```

Using `-a` for script arguments is awkward at best. Consider instead a more natural order:

    dotnet script OPTIONS SCRIPT-PATH SCRIPT-ARGS

All args after the script path should be just passed to the script.

e.g.

    dotnet script --debug ~/scripts/test.csx foo bar baz --verbose

The script will see the args `foo`, `bar`, `baz` & `--verbose`.

PR #23, opened 2016-10-14
Script args at tail for more intuitive usage

atifaziz:
An attempt at addressing #11. I say attempt because while it worked for the script case, it didn't for the eval one & I ran out of my quantum. :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -154,13 +154,13 @@ Hello from foo.csx
 
 ### Passing arguments to scripts
 
-You can pass arguments to the script the following way:
+All arguments after `--` are passed to the script in the following way:
 
 ```
-dotnet script foo.csx -a arg1 -a arg2 -a arg3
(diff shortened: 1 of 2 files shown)


filipw:
This is OK, I don't think `eval` doesn't need arguments IMHO, since the whole code is supplied from the outside anyway.

Can you please update the readme to reflect the arg changes?

atifaziz:
@filipw I've updated the usage in the README. I have also changed the usage so that `--` separates the args passed to `dotnet script` and those passed to the script itself. This has to be done because the command-line args processing library you're using didn't allow `dotnet script` and the script to overload the arguments. `RemainingArguments` would only contain those not processed by `dotnet script`. This means the script could not have its own help (`-h`) or use `-d` to mean something else.

When you're ready to merge (:shipit:), I can squash the changes to a single commit.

filipw:
thanks a lot!

PR #27, opened 2016-10-18
ScriptCompilationContext.Compilation is redundant

atifaziz:
The `Compilation` object is available via `ScriptCompilationContext.Script.GetCompilation` so seems redundant to carry it around.


filipw:
you are right, it's not needed, `Script` will hold a lazy reference

PR #28, opened 2016-10-18
Script args on context should be read-only like rest

atifaziz:
All other properties on `ScriptContext` are read-only so seems inconsistent that the script args should be a modifiable list especially when there's no need ever to modify it. This PR renames `ScriptContext.ScriptArgs` to simply `ScriptContext.Args` & renders it read-only.


filipw:
thanks

PR #47, opened 2017-01-05
added implicit reference to System.ValueTuple

filipw:
Allows us to use code like this:

```
var x = (1, 2);
Console.WriteLine(x.Item1);
Console.WriteLine(x.Item2);
```

This is inline with CSI behavior (also implicitly references `ValueTuple`)


adamralph:
Just an observation: this is a departure from "standard" scripting. I.e. a script which is written for dotnet-script which assumes this reference can no longer be executed with csi.exe.

adamralph:
One day I'll learn to read things properly. Apologies.

filipw:
I appreciate you keeping an eye out on things though - keeps me honest 😄

Issue #55, opened 2017-01-25
Move out any remaining non-CSI compliant components to an opt in package


filipw:
The goal of this project is to provide CSI-like experience on .NET Core.
The remaining non-standard-CSI behavior is the usage of HTTP resolver which was thrown in here initially. 
It should be moved to an external package so that it's usage is an opt in.

seesharper:
@filipw We can close this, right?

PR #63, opened 2017-02-27
moved RemoteFileResolver into a dedicated package

filipw:
Created a package `Dotnet.Script.Extras` for non-CSI compatible features.

Related to #55

### DIFF
--- a/README.md
+++ b/README.md
@@ -179,27 +179,6 @@ Hello from bar.csx
 Hello from foo.csx
 ```
 
-### Referencing an HTTP-based script from a script
-
-Even better, `Dotnet.Script` supports loading CSX references over HTTP too. You could now modify the `foo.csx` accordingly:
-
-```csharp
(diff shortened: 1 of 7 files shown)


README.md
```diff
@@ -224,6 +203,31 @@ dotnet script -d foo.csx -- -d
 ```
 
 will pass the `-d` before `--` to `dotnet script` and enable the debug mode whereas the `-d` after `--` is passed to script for its own interpretation of the argument.
+
+## Extras
+
+Beyond the standard scripting dialect support (compatible with `csi.exe`), `Dotnet.Script` provides some extra features, located in the `Dotnet.Script.Extras` package.
```
atifaziz: So how are the extras enabled now? That information is missing because `#load` using a URL (and the example in the next sub-section) doesn't work anymore.

filipw: at the moment they are not. 
Once #55 is resolved there will be a discovery process in place.
Note that this is not released at the moment anywhere, so nothing is broken.

PR #64, opened 2017-02-27
Paket PoC

filipw:
Paket integration done using `#r` extensibility in Roslyn. 

Given a script like this:

```
#r "paket: nuget Newtonsoft.Json"

using System;
using Newtonsoft.Json;

class Foo
{
    public string Bar { get; set; }
}

Console.WriteLine(JsonConvert.SerializeObject(new Foo { Bar = "hi" }));
```

This happens:

screenshot 2017-02-27 16 46 41

### DIFF
--- a/Dotnet.Script.sln
+++ b/Dotnet.Script.sln
@@ -9,6 +9,10 @@ Project("{8BB2217D-0F2D-49D1-97BC-3654ED321F3B}") = "Dotnet.Script.Core", "src\D
 EndProject
 Project("{8BB2217D-0F2D-49D1-97BC-3654ED321F3B}") = "Dotnet.Script.Extras", "src\Dotnet.Script.Extras\Dotnet.Script.Extras.xproj", "{BCBB3155-4463-4748-8032-EF82AE297CE1}"
 EndProject
+Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "examples", "examples", "{61E3CEE0-3954-486F-AD1A-E5EE66DEED62}"
+EndProject
+Project("{8BB2217D-0F2D-49D1-97BC-3654ED321F3B}") = "Dotnet.Script.Extras.Demos", "examples\Dotnet.Script.Extras.Demos\Dotnet.Script.Extras.Demos.xproj", "{522E8334-72C1-43F4-AFF3-3D504EE46DFE}"
+EndProject
 Global
 	GlobalSection(SolutionConfigurationPlatforms) = preSolution
 		Debug|Any CPU = Debug|Any CPU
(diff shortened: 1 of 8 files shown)


forki:
wow cool

smoothdeveloper:
Amazing @filipw :)

atifaziz:
> this is very dirty at the moment

So what would a clean implementation look like? 😄 I think what Paket needs is a command that simply resolves and lists those references that it generates in load scripts, enabling it to be programmatically invoked and the output harnessed.

forki:
What's wrong with the load script? That's a very practical approach -
especially if you want to debug if something goes wrong.

Am 11.03.2017 12:44 nachm. schrieb "Atif Aziz" <notifications@github.com>:

atifaziz:
@forki Nothing is wrong with the script per se but it's not the same scenario. I'm talking about an app, like this one here, that wants to use Paket as a package and dependency manager for its own purpose and something that's more or less transparent to the user. This would avoid hard-coding a couple of things like [the script file name](https://github.com/filipw/dotnet-script/pull/64/files#diff-a5f7ec6e709fcbdac643027555aad9afR64) and [shape of references](https://github.com/filipw/dotnet-script/pull/64/files#diff-a5f7ec6e709fcbdac643027555aad9afR71) (both of which could change with future versions) and instead directly harness the standard output from a hypothetical subcommand like `list-refs`. For example, `git status` gives you a user-friendly output whereas [`git status --porcelain`](https://git-scm.com/docs/git-status#git-status---porcelainltversiongt) emits in versioned machine-readable format for script/program consumption. The latter allows programs to use other programs as larger building blocks, without resorting to use of libraries and bindings in the language of consuming app.

forki:
Of course we could emit the same information via command line in paket.
That would be pretty easy (and a good way to solve trust issues since load
files could potentially do big harm.), but the implementation of this in
visualfsharp shows that load files also can work as cache. This is
important for tooling scenarios in the IDE (where we need to resolve on
every keystroke). Basically we don't call paket at all when the script is
already there and the temporary deps file didn't change. Of course you can
do similar stuff in an IDE layer, but probably not easily across different
sessions.

Am 11.03.2017 13:45 schrieb "Atif Aziz" <notifications@github.com>:

atifaziz:
> Of course we could emit the same information via command line in paket. That would be pretty easy (and a good way to solve trust issues since load files could potentially do big harm.)

@forki Cool, so you would consider adding this if I open an issue in the Paket repo for tracking?

forki:
No but I would probably merge a pull request ;-)

Am 11.03.2017 2:26 nachm. schrieb "Atif Aziz" <notifications@github.com>:

atifaziz:
> No but I would probably merge a pull request ;-)

I thought you said it would be pretty easy, :wink: but okay, fair enough. If I don't get the time myself (guessing it'll take time to board on the project and “get into it”), I might tempt someone (@gmpl?) to look into it. I still think opening an issue for tracking and discussion of the feature may not hurt meanwhile, instead of riding sideways on this thread.

forki:
Yes please open an issue. There is a good chance someone will pick it up. I
was just kidding a bit.

Am 11.03.2017 3:02 nachm. schrieb "Atif Aziz" <notifications@github.com>:

filipw:
I'll close this as it got out of date and will reopen a new one applying the learnings that happened over the past year

PR #66, opened 2017-03-02
Updated to csproj / latest dotnet tooling

### LINKED ISSUE #58: Migrate to csproj
Time to move the project to csproj

filipw:
Fixes #58 

The only supported IDEs are now Visual Studio 2017 and VS Code with C# Extension 1.7.0+


Dotnet.Script.sln
```diff
@@ -1,32 +1,59 @@
...
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Dotnet.Script.Core", "src\Dotnet.Script.Core\Dotnet.Script.Core.csproj", "{684FEFEE-451B-4E68-B662-C16E3A7DA794}"
 EndProject
-Project("{8BB2217D-0F2D-49D1-97BC-3654ED321F3B}") = "Dotnet.Script.Extras", "src\Dotnet.Script.Extras\Dotnet.Script.Extras.xproj", "{BCBB3155-4463-4748-8032-EF82AE297CE1}"
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Dotnet.Script.Extras", "src\Dotnet.Script.Extras\Dotnet.Script.Extras.csproj", "{BCBB3155-4463-4748-8032-EF82AE297CE1}"
 EndProject
 Global
 	GlobalSection(SolutionConfigurationPlatforms) = preSolution
 		Debug|Any CPU = Debug|Any CPU
+		Debug|x64 = Debug|x64
+		Debug|x86 = Debug|x86
```
adamralph: Is this something you chose to do, or is it part of the upgrade?

global.json
```diff
@@ -1,6 +1,6 @@
 {
-  "projects": [ "src" ],
-  "sdk": {
-    "version": "1.0.0-preview2-1-003177"
-  }
+    "projects": [ "src" ],
+    "sdk": {
+            "version": "1.0.0-rc4-004771"
+    }
```
adamralph: We still need global.json in the new world?

Dotnet.Script.sln
filipw: part of the tool

global.json
filipw: yes you want to pin to specific tooling version as there is no guarantee it will not break with future releases. 

plus if you are missing the right sdk version, global.json ensures we don't attempt to build with an older one and explode

PR #78, opened 2017-04-05
Started using Roslyn 3.0.0-beta (scripting)

seesharper:
Removed the DebugScriptRunner.
Think we still need to specify UTF8


NuGet.Config
```diff
@@ -0,0 +1,6 @@
+<configuration>
+<packageSources>
+ <add key="nuget.org" value="https://api.nuget.org/v3/index.json" protocolVersion="3" />
+ <add key="Roslyn nightly builds" value="https://dotnet.myget.org/F/roslyn/api/v3/index.json" />
```
filipw: we should not be using the nightly build feed. I'd rather internalize the DLLs so that we can publish a nuget package

src/Dotnet.Script/Program.cs
```diff
@@ -81,16 +81,16 @@ private static void RunScript(string file, string config, bool debugMode, IEnume
             }
 
             var directory = Path.IsPathRooted(file) ? Path.GetDirectoryName(file) : Path.GetDirectoryName(Path.Combine(Directory.GetCurrentDirectory(), file));
-            var sourceText = SourceText.From(new FileStream(file, FileMode.Open));
-            var context = new ScriptContext(sourceText, directory, config, args, file);
+            var sourceText = SourceText.From(new FileStream(file, FileMode.Open),Encoding.UTF8);
```
filipw: you should not hardcode the encoding. `SourceText.From(stream)` will internally try to detect encoding using BOM and if it's not found, it will fallback to UTF8 automatically

src/Dotnet.Script.Core/ScriptCompiler.cs
```diff
@@ -45,15 +47,15 @@ public ScriptCompiler(ScriptLogger logger)
...
+            var opts = ScriptOptions.Default.AddImports(ImportedNamespaces)
+                .AddReferences(ReferencedAssemblies)
+                .WithSourceResolver(SourceFileResolver.Default)
+                .WithMetadataResolver(ScriptMetadataResolver.Default)
+                .WithEmitDebugInformation(context.DebugMode);            
 
             if (!string.IsNullOrWhiteSpace(context.FilePath))
             {
-                opts = opts.WithFilePath(context.FilePath);
+                opts = opts.WithFilePath(context.FilePath).WithFileEncoding(Encoding.UTF8);
```
filipw: same here, encoding will not be needed if you change here https://github.com/filipw/dotnet-script/blob/master/src/Dotnet.Script.Core/ScriptCompiler.cs#L108 to use the new stream-based API https://github.com/dotnet/roslyn/blob/master/src/Scripting/CSharp/CSharpScript.cs#L44

filipw:
thanks a lot! just have a few comments

NuGet.Config
seesharper: By internalizing you mean that we should keep the Roslyn dll's in the repo?

seesharper: Regardless how to distribute, we still ned to bring in the 3.0.0 dll's from somewhere before we create the NuGet package?

filipw: yeah in scriptcs I just copied the DLLs into the `lib` folder and included them into the created nuget package https://github.com/scriptcs/scriptcs/commit/44f666e82c537fcdde87529eb286c9a7558ae26d

Alternatively, but I'd like to avoid it, we can take Roslyn 3 from myget and publish on nuget under a different name and reference that 😎

seesharper: Going for the lib folder approach. Do you know what we have to do in csproj to include these files in the NuGet package?

filipw: something like this should work

```
  <ItemGroup>
    <Content Include="$(OutputPath)\Microsoft.CodeAnalysis.CSharp.Scripting.dll">
      <PackagePath>lib/netstandard1.6/</PackagePath>
      <Pack>true</Pack>
    </Content>
  </ItemGroup>
```

src/Dotnet.Script.Extras/Dotnet.Script.Extras.csproj
```diff
@@ -11,10 +11,26 @@
     <GenerateAssemblyConfigurationAttribute>false</GenerateAssemblyConfigurationAttribute>
     <GenerateAssemblyCompanyAttribute>false</GenerateAssemblyCompanyAttribute>
     <GenerateAssemblyProductAttribute>false</GenerateAssemblyProductAttribute>
+    <Version>0.2.0</Version>
```
filipw: this is not needed, there is already a `<VersionPrefix />`

src/Dotnet.Script.Core/Dotnet.Script.Core.csproj
```diff
@@ -18,12 +18,35 @@
     <GenerateAssemblyConfigurationAttribute>false</GenerateAssemblyConfigurationAttribute>
     <GenerateAssemblyCompanyAttribute>false</GenerateAssemblyCompanyAttribute>
     <GenerateAssemblyProductAttribute>false</GenerateAssemblyProductAttribute>
+    <Version>0.11.0</Version>
```
filipw: this is not needed, there is already a `<VersionPrefix />`

filipw:
looks good, can you please just fix the version and we merge it

src/Dotnet.Script.Core/Dotnet.Script.Core.csproj
adamralph: The version properties are confusing.

Related:
- https://github.com/Microsoft/msbuild/issues/1952
- https://twitter.com/adamralph/status/850238043709521924

seesharper: https://github.com/filipw/dotnet-script/pull/78/commits/176cc676770a3e10c76278403c460e813eaaf422

adamralph: @seesharper sure, I see you've fixed it. Mine was just a tangential comment.

filipw:
thanks a lot

PR #82, opened 2017-06-02
switch from InteractiveScriptGlobals to CommandLineScriptGlobals

filipw:
CSI uses `CommandLineScriptGlobals` in script execution context, `InteractiveScriptGlobals` is in REPL mode only.

### DIFF
--- a/src/Dotnet.Script.Core/ScriptRunner.cs
+++ b/src/Dotnet.Script.Core/ScriptRunner.cs
@@ -19,12 +19,12 @@ namespace Dotnet.Script.Core
 
         public Task<TReturn> Execute<TReturn>(ScriptContext context)
         {
-            var globals = new InteractiveScriptGlobals(Console.Out, CSharpObjectFormatter.Instance);
+            var globals = new CommandLineScriptGlobals(Console.Out, CSharpObjectFormatter.Instance);
 
             foreach (var arg in context.Args)
                 globals.Args.Add(arg);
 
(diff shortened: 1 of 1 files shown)


adamralph:
> CSI uses CommandLineScriptGlobals in script execution context, InteractiveScriptGlobals is in REPL mode only.

Wow, that's ugly! You can imagine a command to save your current interactive session as a script, so you can run it again via the command line, but when you try to do so... BOOM! You used script globals in your session which aren't available to the command line. 😧

filipw:
the only difference between the two is that `InteractiveScriptGlobals` exposes search paths as global properties.

anyway, even if you have a session saved (which is a hypothetical functionality) for later use, you'd still replay it with the interactive mode - evaluating line by line (or rather block by block) not as continuous script. And in that case, you could always ensure to pick the correct host object

adamralph:
Yeah, I suppose so. So I guess it would need some other file extension to distinguish? Or I guess this could be part of the "script dialects" we've been talking about, i.e. something inside the file which tells the runner what globals etc. to provide.

filipw:
> So I guess it would need some other file extension to distinguish?

yes - because a REPL "session export" wouldn't be a CSX file, as the semantics of re-rerunning a REPL "session" and running a script file are different. For example in REPL you can redefine same members or types over and over, which is invalid in a single CSX.

adamralph:
Ah yes, of course.

PR #91, opened 2017-08-04
Incorporate Inline Nuget Resolver

filipw:
@seesharper can you have a look if it makes sense?

### DIFF
--- a/src/Dotnet.Script.Core/Dotnet.Script.Core.csproj
+++ b/src/Dotnet.Script.Core/Dotnet.Script.Core.csproj
@@ -21,6 +21,7 @@
   </PropertyGroup>
 
   <ItemGroup>
+    <PackageReference Include="Dotnet.Script.NuGetMetadataResolver" Version="2.0.3" />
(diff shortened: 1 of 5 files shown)


seesharper:
Looks good to me, @filipw 👍 

Just one question 

How do we picture the install experience when there is no project.json file to begin with where the dotnet-script package is referenced? Could we use Chocolatey on Windows to bring the dotnet-script.cmd into path?

It would be really nice if we could somehow install globally and be able to just execute dotnet-script [scriptfile] from any directory.

filipw:
> It would be really nice if we could somehow install globally and be able to just execute dotnet-script [scriptfile] from any directory.

exactly. so there is a little known trick here as well. 
if you have `dotnet-foo` in your PATH, then .NET CLI will discover it and you allow you to call `dotnet foo` command, which is what I was trying to achieve 😄 

so if you have `dotnet-script` in your path, we can maintain the `dotnet script` experience

filipw:
I think using choco will be definitely needed, for now I think I'll just publish the binaries here as a release

PR #94, opened 2017-08-05
A few fixes

filipw:
- point `SourceFileResolver` to current working directory, this way we won't fail on load resolution when running dotnet exec from other locations
 - suppress CS1701 and related diagnostics (see: https://github.com/dotnet/roslyn/issues/5501)
 - removed dead DLLs
 - when using `scriptProjectProvider`, use `netcoreapp1.0` for the time being


seesharper:
Nice work 👍  I'm merging this if that is okay with you?

filipw:
thanks

PR #96, opened 2017-08-07
Some more bug fixes - this time to make sure nuget resolver installs correctly

### LINKED ISSUE #89: Support .NETCoreApp 1.1
I wonder how much effort is needed to support .NETCoreApp 1.1?

filipw:
- bump everything to netcoreapp1.1 (fixes #89)
 - removed package fallback (looks like this caused https://github.com/seesharper/Dotnet.Script.NuGetMetadataResolver/issues/2)
 - removed implicit net standard reference (not needed)
 - removed unneeded csproj config (not used)


filipw:
This PR also fixes bugs related to relative path handling. 
At the moment, requests like `dotnet script ../../foo/bar.csx` will have incorrect working directory which will result in bad resolution of i.e. load and r directives.

PR #97, opened 2017-08-10
Added support for csharp 7.1

filipw:
Roslyn has this change already merged in, but it will take a while before it's released as stable on nuget.
In the meantime, we can set latest C# version with reflection.

Omnisharp also already has support for latest C# in scripts (see https://github.com/OmniSharp/omnisharp-roslyn/pull/935)


filipw:
correct

PR #98, opened 2017-08-12
Feature/testing

seesharper:
Added initial test project that allows us to test script execution with different test script fixtures. We will probably face users having trouble with certain scripts and then we can simply put them here for trouble shooting and regression testing.


filipw:
great, thanks

PR #104, opened 2017-08-17
Fixed path to dotnet-script.dll according to debug/release configuration

seesharper:
Discovered this when I tried to run the tests in release mode


filipw:
thanks

PR #105, opened 2017-08-17
Fixed exit code when script fails

### LINKED ISSUE #103: ExitCode 0 even when script fails
At the moment we basically swallow the exception from the script.
https://github.com/filipw/dotnet-script/blob/master/src/Dotnet.Script.Core/ScriptRunner.cs#L48
I do have a fix for this although it would be best to apply it after #100 is merged

seesharper:
Fixes #103


filipw:
OK great, appveyor now triggers automatically

PR #106, opened 2017-08-17
Feature/appveyor

seesharper:
This PR adds CI integration with AppVeyor. All you need to do is to add this repo to your AppVeyor projects. I assume that you would want to "own" the project on AppVeyor. 

The interesting part here is that we build dotnet-script using dotnet-script :) 
I figured that a little dogfooding  wouldn't hurt so I basically took what I have to build LightInject and simplified it a lot. It still represents a pretty good use case for C# scripting.

I'll will take a look at Travis next, so that we can build and verify all test fixtures on all platforms.


filipw:
Excellent! I added the project to the appveyor account, for some reason it's not picking up the open PR at the moment, but let's merge and see how it works off master. Thanks again!

filipw:
perfect! https://ci.appveyor.com/project/filipw/dotnet-script/build/1.0.2

PR #108, opened 2017-08-22
Removed dotnet command suffix (.exe) and changed to lowercase

seesharper:
The dotnet command need to be lower case and without the .exe suffix in order to run on *nix


filipw:
merci

PR #111, opened 2017-08-28
Downgrade tool to netcoreapp1.0

devlead:
.NET Core tools are still `netcoreapp1.0` so using *dotnet-script* as a tool today will fail during restore

Example this csproj 
```xml
 <PropertyGroup>
   <TargetFramework>netcoreapp1.1</TargetFramework>
 </PropertyGroup>
   <ItemGroup>
      <DotNetCliToolReference Include="Dotnet.Script" Version="0.12.0-beta" />
    </ItemGroup>
</Project>
```
will fail wit the error
```diff
+ > dotnet restore .\hello.csproj --source C:\temp\dev\gh\devlead\dotnet-script\nuget
-  error NU1202: Package Dotnet.Script 0.12.0-beta is not compatible with netcoreapp1.0 (.NETCoreApp,Version=v1.0).
-  Package Dotnet.Script 0.12.0-beta supports: netcoreapp1.1 (.NETCoreApp,Version=v1.1)
```

Adjusting the tool target framework seems to do the trick (tested with dotnet build/test/pack to local folder and that folder used as source)
```diff
+ > dotnet restore .\hello.csproj --source C:\temp\dev\gh\devlead\dotnet-script\nuget
+   Restoring packages for C:\temp\dotnet-script-test\hello.csproj...
+   Restoring packages for C:\temp\dotnet-script-test\hello.csproj...
+   Installing Dotnet.Script 0.13.0.
+   Generating MSBuild file C:\temp\dotnet-script-test\obj\hello.csproj.nuget.g.props.
+   Restore completed in 814,17 ms for C:\temp\dotnet-script-test\hello.csproj.
+   Restore completed in 862,33 ms for C:\temp\dotnet-script-test\hello.csproj.
+ C:\temp\dotnet-script-test> dotnet script .\hello.csx
+ Hello
```

### DIFF
--- a/src/Dotnet.Script/Dotnet.Script.csproj
+++ b/src/Dotnet.Script/Dotnet.Script.csproj
@@ -4,7 +4,7 @@
     <Description>Dotnet CLI tool allowing you to run C# (CSX) scripts.</Description>
     <VersionPrefix>0.13.0</VersionPrefix>
     <Authors>filipw</Authors>
-    <TargetFramework>netcoreapp1.1</TargetFramework>
+    <TargetFramework>netcoreapp1.0</TargetFramework>
     <DebugType>portable</DebugType>
     <AssemblyName>dotnet-script</AssemblyName>
     <OutputType>Exe</OutputType>
@@ -18,7 +18,7 @@
     <PackageTargetFallback>$(PackageTargetFallback);dotnet5.6;portable-net45+win8</PackageTargetFallback>
     <GenerateAssemblyConfigurationAttribute>false</GenerateAssemblyConfigurationAttribute>
     <GenerateAssemblyCompanyAttribute>false</GenerateAssemblyCompanyAttribute>
-    <GenerateAssemblyProductAttribute>false</GenerateAssemblyProductAttribute>      
+    <GenerateAssemblyProductAttribute>false</GenerateAssemblyProductAttribute>
   </PropertyGroup>
 
   <ItemGroup>
@@ -29,7 +29,7 @@
       <CopyToOutputDirectory>Always</CopyToOutputDirectory>
     </Content>
   </ItemGroup>
-     
+
   <ItemGroup>
     <ProjectReference Include="..\Dotnet.Script.Core\Dotnet.Script.Core.csproj" />
   </ItemGroup>


seesharper:
We are actually looking into targeting netcoreapp2.0 in an effort to solve a couple of runtime dependency problems we are seeing on Linux. The way we will integrate with the dotnet cli is by putting the dotnet-script.cmd/dotnet-script.sh on the PATH rather than explicitly having the tool reference. References to NuGet packages are being done inline like explained in these release notes.

https://github.com/filipw/dotnet-script/releases/tag/0.11.0-beta

devlead:
Ah I see, so bootrapping will not use the dotnet cli for more than execution of the tool if already in place in path?

I get the nuget references for script dependencies, but how do I obtain the dotnet cli command script without using the .NET CLI? Is the intent here to use something custom?

seesharper:
> Ah I see, so bootrapping will not use the dotnet cli for more than execution of the tool if already in place in path?

Yes, we integrate using PATH-based extensibility according to these docs 
https://docs.microsoft.com/en-us/dotnet/core/tools/extensibility

> I get the nuget references for script dependencies, but how do I obtain the dotnet cli command script without using the .NET CLI? Is the intent here to use something custom?

Once the dotnet-script.cmd/sh file is in the path you can invoke it just as you would with any other cmd/sh file or you can invoke it through the dotnet cli (dotnet script).

devlead:
I'll go ahead and close this PR then. 

So basically curl the nuget package, unzip and add to path is the recommended way?

filipw:
Hi Mattias, thanks a lot for the PR.
The error you are seeing is only visible if you use 1.0.x SDK for .NET Core. If you use .NET Core 2.0.0 SDK the problem is already fixed.

By the way, there is a way to force the external tool to run under different CLI runtimes. Here is the example
https://github.com/dotnet/cli/tree/rel/1.0.0/TestAssets/TestPackages/dotnet-prefercliruntime 
Perhaps it's something we should do on this package, although I'm not convinced it's necessary (yet), especially as we'll like switch the whole thing to .NET Core 2.

The mentioned above way of integrating into .NET CLI will be the recommended way going forward, yes. Now, it won't work by grabbing the nuget package, because you will need to have things like runtimeconfig resolved and so on - that happens normally on dotnet restore. I personally think the easiest would be to point to the github release and unzip that.

devlead:
Perhaps a dotnet-script.portable Chocolatey package it could contain everything the release does, then you could `choco install` or just curl&unzip, and that way still get the advantage of NuGet server side bits and pieces.

PR #116, opened 2017-08-30
Added version option

seesharper:
This PR adds support for displaying the version number. 
I've added the beta suffix to the actual version number for now.

```
dotnet script -v|--version
0.13.0-beta
```

### DIFF
--- a/src/Dotnet.Script/Program.cs
+++ b/src/Dotnet.Script/Program.cs
@@ -43,7 +43,7 @@ namespace Dotnet.Script
 
             app.HelpOption("-? | -h | --help");
 
-            app.VersionOption("-v | --version", () => $"{typeof(Program).GetTypeInfo().Assembly.GetName().Version.ToString(3)}-beta");
+            app.VersionOption("-v | --version", GetVersionInfo);
 
             app.Command("eval", c =>
             {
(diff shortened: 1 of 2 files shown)


src/Dotnet.Script/Program.cs
```diff
@@ -42,6 +43,8 @@ private static int Wain(string[] args)
 
             app.HelpOption("-? | -h | --help");
 
+            app.VersionOption("-v | --version", () => $"{typeof(Program).GetTypeInfo().Assembly.GetName().Version.ToString(3)}-beta");
+
```
maximpashuk: I think you should use value of AssemblyInformationalVersionAttribute instead of assembly version.

So there will no need to hardcode '-beta' in code, since it is subject to change in future.

adamralph: 👍 x 💯

filipw: this is a good suggestion. Something like this: `typeof(Program).GetTypeInfo().Assembly.GetCustomAttribute<AssemblyInformationalVersionAttribute>().InformationalVersion;` will indeed automatically print `0.13.0-beta` when built with the beta suffix, and `0.13.0` when built without it.

seesharper: Fixing it now :)

seesharper: I've included the VersionSuffix in the csproj file. Makes it easier to read the version info when building choco packages which is what I am doing now  👍

filipw:
perfect - thanks

PR #119, opened 2017-09-05
added build.ps1 and build.sh

filipw:
Added single build entry point build files.
Also noticed some issues we have with current directory, solved it by allowing to pass in the root path.


seesharper:
The Travis push build seems stuck. I'm merging this since the pr build completed successfully

PR #124, opened 2017-09-12
Bugfix/chocolatey

seesharper:
We should automate building verification.txt, but let's just do it manually for now


seesharper:
We need to take this before #122  :)

filipw:
Merci!

PR #125, opened 2017-09-12
renamed choco package to dotnet-script

seesharper:
This comes in a little late, but I suddenly realized that the choco package was named `dotnet.script`. IMHO a better name would be `dotnet-script` since that is sort of the name of the "product". It is actually recommended to use hyphen rather than dots. 

I simply want to avoid the possible confusion of having a different name for the product and the choco package. 

https://github.com/chocolatey/choco/wiki/CreatePackages#naming-your-package

On a side note, it is strange that the validation process did not complain about this. 

I should have thought about this before pushing to Chocolatey, but we simply unlist dotnet.script and push dotnet-script instead. It is now or never :)

### DIFF
--- a/build/Build.csx
+++ b/build/Build.csx
@@ -9,7 +9,7 @@ var root = Args.FirstOrDefault() ?? "..";
 
 DotNet.Build($"{root}/src/Dotnet.Script");
 DotNet.Build($"{root}/src/Dotnet.Script.Tests");
-// DotNet.Test($"{root}/src/Dotnet.Script.Tests");
+DotNet.Test($"{root}/src/Dotnet.Script.Tests");
 DotNet.Publish($"{root}/src/Dotnet.Script");
 
 if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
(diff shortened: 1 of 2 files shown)


filipw:
it's really only called `dotnet-script`, because the dotnet CLI extensibility mandates such naming format. 
The nuget package is still called `dotnet.script` so I don't think there is no problem in chocolatey package being called the same

seesharper:
Yes, I understand the rationale behind the name :) Just wanted to make sure that we are all set on naming before we go "public" with this. One more thing to consider. Would it ever be a possibility that we make the choco package available from NuGet as well? In that case, we need another name so that it does not conflict with the dotnet.script library package. We also get the situation where given the same package identity, we get two different things depending on where it was resolved from. What if somebody puts the chocolatey feed into their NuGet.Config file for some reason? Sorry for the nitpicking, but I have been bitten so many times by related naming issues :) If we keep `dotnet.script` as the name of the choco package, it is at least super important that we refer to the product as `dotnet.script' everywhere in the documentation.

seesharper:
Had the day to think about this and figured it will be okay to stick with `dotnet.script` as the the package name.  Updated the readme here (https://github.com/filipw/dotnet-script/tree/feature/readme) so that we refer to it as `dotnet script` which makes sense since that is what people are going to execute:)

PR #131, opened 2017-09-23
fixes #129

### LINKED ISSUE #129: Could not load Newtonsoft.Json assembly
I'm using .Net Core 1.1.4: https://github.com/dotnet/core/blob/master/release-notes/download-archives/1.1.4-download.md

I downloaded 0.13 of `dotnet-script` and I can't seem to figure out why I'm getting the error below. I've tried including a different nuget reference of `#r "nuget:Newtonsoft.Json, 10.0.0"`.
```
Could not load file or assembly 'Newtonsoft.Json, Version=10.0.0.0, Culture=neutral, PublicKeyToken=30ad4fe6b2a6aeed'. Could not find or load a specific file. (Exception from HRESULT: 0x80131621)
   at Auth0.ManagementApi.Clients.ConnectionsClient.GetAllAsync(String strategy, String fields, Boolean includeFields, String name)
   at Submission#0.<<Initialize>>d__0.MoveNext() in /Users/saqibrokadia/invio/scripts/auth0/auth/helloworld.csx:line 9
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at Microsoft.CodeAnalysis.Scripting.ScriptExecutionState.<RunSubmissionsAsync>d__9`1.MoveNext()
Could not load file or assembly 'Newtonsoft.Json, Version=10.0.0.0, Culture=neutral, PublicKeyToken=30ad4fe6b2a6aeed'. Could not find or load a specific file. (Exception from HRESULT: 0x80131621)
```

I think this has to do with dependency resolution of nuget, but feels like I'm pretty stuck here and not sure what else to try. I've read through other issues in this repository, but nothing seems to fix this.
```
#! "netcoreapp1.0"
#r "~/.nuget/packages/Newtonsoft.Json/10.0.3/lib/netstandard1.0/Newtonsoft.Json.dll"
#r "nuget:Auth0.ManagementApi, 4.4.0"

using Auth0.ManagementApi;

var client = new ManagementApiClient("token", new Uri("https://foo.auth0.com/api/v2/"));

Console.WriteLine(await client.Connections.GetAllAsync("auth0"));
```

seesharper:
This PR fixes issue #129 . More specifically we now only try to redirect an assembly if the request assembly version is different from what we have already resolved as a runtime dependency.

We will receive more of these issues and I think the only sane way to go about this is to create tests that pretty much does exactly the same as the example provided in the issues. It would take us forever if we should try to recreate the particularities of a given package. If a package gets unlisted for some reason, we will know about it and can take action accordingly.  👍

### DIFF
--- a/src/Dotnet.Script.Tests/ScriptExecutionTests.cs
+++ b/src/Dotnet.Script.Tests/ScriptExecutionTests.cs
@@ -34,10 +34,18 @@ namespace Dotnet.Script.Tests
             Assert.Contains("Connection successful", result.output);
         }
         
+        [Fact]
         public static void ShouldReturnNonZeroExitCodeWhenScriptFails()
         {
             var result = Execute(Path.Combine("Exception", "Error.csx"));
             Assert.NotEqual(0, result.exitCode);
+        }
+
+        [Fact]
+        public static void ShouldHandleIssue129()
+        {
+            var result = Execute(Path.Combine("Issue129", "Issue129.csx"));    
+            Assert.Contains("Bad HTTP authentication header", result.output);
         }
 
         private static (string output, int exitCode) Execute(string fixture)
--- a/build/Build.csx
+++ b/build/Build.csx
@@ -9,7 +9,7 @@ var root = Args.FirstOrDefault() ?? "..";
 
 DotNet.Build($"{root}/src/Dotnet.Script");
 DotNet.Build($"{root}/src/Dotnet.Script.Tests");
-// DotNet.Test($"{root}/src/Dotnet.Script.Tests");
+DotNet.Test($"{root}/src/Dotnet.Script.Tests");
 DotNet.Publish($"{root}/src/Dotnet.Script");
 
 if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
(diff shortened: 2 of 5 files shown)


seesharper:
The tests are now back in business. 👍

src/Dotnet.Script.Tests/ScriptExecutionTests.cs
```diff
@@ -34,12 +34,20 @@ public void ShouldHandlePackageWithNativeLibraries()
...
         public static void ShouldReturnNonZeroExitCodeWhenScriptFails()
         {
             var result = Execute(Path.Combine("Exception", "Error.csx"));
             Assert.NotEqual(0, result.exitCode);
         }
 
+        [Fact]
+        public static void ShouldHandleIssue129()
+        {
+            var result = Execute(Path.Combine("Issue129", "Issue129.csx"));
```
filipw: nice 👍

filipw:
Beautiful

PR #136, opened 2017-10-05
Bug/issue 135

### LINKED ISSUE #135: Passing arguments to dotnet script at the command line doesn't work.
I have the following foo.csx file (from the example in [README.md](https://github.com/filipw/dotnet-script/blob/master/README.md)):

```C#
#! "netcoreapp1.1"
#r "nuget:NetStandard.Library,1.6.1"

foreach (var arg in Args)
{
    Console.WriteLine(arg);
}
```
When I try to run dotnet script at the command line I see the following:
```shell
>  dotnet script .\foo.csx -- arg1 arg2 arg3
Couldn't find file '.\foo.csx -- arg1 arg2 arg3'
```
I've tried this from cmd and PowerShell.

I am running version 0.13.0 of dotnet-script on Windows 10.

seesharper:
This PR fixes #135 

It now works like advertised in the readme :)

### DIFF
--- a/src/Dotnet.Script/Program.cs
+++ b/src/Dotnet.Script/Program.cs
@@ -40,11 +40,14 @@ namespace Dotnet.Script
         }
 
         private static int Wain(string[] args)
-        {
-            var app = new CommandLineApplication(throwOnUnexpectedArg: false);
-            var file = app.Argument("script", "Path to CSX script");
+        {            
+            var app = new CommandLineApplication(throwOnUnexpectedArg: false);           
+            var file = app.Argument("script", "Path to CSX script");            
             var config = app.Option("-conf |--configuration <configuration>", "Configuration to use. Defaults to 'Release'", CommandOptionType.SingleValue);
             var debugMode = app.Option(DebugFlagShort + " | " + DebugFlagLong, "Enables debug output.", CommandOptionType.NoValue);
+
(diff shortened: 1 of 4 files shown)


src/Dotnet.Script/Program.cs
```diff
@@ -60,7 +63,7 @@ private static int Wain(string[] args)
                 {
                     if (!string.IsNullOrWhiteSpace(code.Value))
                     {
-                        await RunCode(code.Value, config.HasValue() ? config.Value() : "Release", debugMode.HasValue(), app.RemainingArguments, cwd.Value());
+                        await RunCode(code.Value, config.HasValue() ? config.Value() : "Release", debugMode.HasValue(), app.RemainingArguments.Concat(argsAfterDoubleHypen), cwd.Value());
```
filipw: how did it work before? I mean, even the build script uses this https://github.com/filipw/dotnet-script/blob/master/build/Build.csx#L8

filipw:
Thanks a lot - but I'm baffled how that happened, cause it used to work fine.

src/Dotnet.Script/Program.cs
filipw: ugh I is it because the CMD script wraps everything into a quote automatically?
I believe everything works fine when going through the DLL manually, but not when going through CMD script

seesharper:
Are we good here, @filipw ?

filipw:
👍 thanks!

PR #138, opened 2017-10-06
Quote path in cmd in case of spaces

atifaziz:
Since we are on the topic of improving the `cmd`, as in 27b179626b1d8d730aac66994690bccbd551d9dd, this PR quotes the path to `dotnet-script.dll` correctly in the event that there's a directory in the path with a space in its name, like `Program Files`.


filipw:
thanks!

PR #140, opened 2017-10-10
Feature/netcore20

seesharper:
This PR adds support for running `dotnet.script` as a `netcoreapp2.0` application. It is a big update as this PR also lays the ground for resolving the compilation dependencies from OmniSharp.Scripting.

There are two new projects/packages being produced now
 * Dotnet.Script.DependencyModel.NuGet
 * Dotnet.Script.DependencyModel.NuGet.NuGet

Once we publish 0.14.0 and these packages are out we can create a PR (have this prepared already) for OmniSharp.Script that takes us away from the old `project.json` once and for all :)

### Dotnet.Script.DependencyModel

Provides an API to resolve runtime and compilation dependencies for csproj-based scripts. 

This library has no dependencies to other `dotnet.script` libraries or other third-party packages, which means that it can be used by all script runners that needs to resolve runtime dependencies.

In addtion to being compiled for `netstandard2.0`,  it is also compiled for `net46` so that it can be used from `OmniSharp.Scripting` 

### Dotnet.Script.DependencyModel.NuGet

This package contains ONLY the `NuGetMetadataReferenceResolver` that is used to handle inline NuGet package references. 

```
#r "nuget: AutoMapper, 9.1.0"
```

This `MetadataReferenceResolver` does not really do anything besides allowing the actual NuGet reference and returning an assembly reference. 

### Prerequisites

The `Dotnet 2.0 sdk` will have to be installed since we use `dotnet restore` under the hood to provide a `DependencyContext` from which we resolve runtime and compilation dependencies. On the execution side ,this is not a problem since the user would have to have the `dotnet cli` installed anyway to execute `dotnet.script`. On the OmniSharp side, relying on the `dotnet cli` might sound a little more questionable. But since `dotnet.script` is the only script runner that actually understands either an explicit `csproj` file or a `csproj` file generated from script NuGet references, we should be good also on the OmniSharp side. 
We could maybe have used the NuGet.Commands package which is the package that the `dotnet cli` uses under the cover, but the documentation here is virtually non-existent in addition to that it brings in a lot of dependencies that we don't necessarily want to have in OmniSharp.

### Logging

Instead of relying on a logging framework, we simply provide a delegate that can be used from the consuming library to forward log messages to the actual logging framework. 
This is to done to reduce the number of third party dependencies needed by `Dotnet.Script.DependencyModel`

Example:

```c#
LogFactory logFactory = type =>
{
    // Create actual logger here
    return ((level, message) =>
    {
        // Forward log message to actual logger here.
    });   
```

> Note: If we pass `LogLevel.Debug`, we are being verbose and it should probably be logged as a debug statement. Otherwise it should appear in the standard log output. 

### Resolving runtime dependencies (Execution)

```c#
LogFactory logFactory = type =>
{
    // Create actual logger here
    return ((level, message) =>
    {
        // Forward log message to actual logger here.
    });   

var resolver  = new RuntimeDependencyResolver(logFactory);
var runtimeDependencies = resolver.GetDependencies(targetDirectory);
```

### Resolving packages with native/unmanaged dependencies

Some NuGet packages ship with native/unmanaged libraries. An example would be the [Microsoft.Data.SQLite](https://www.nuget.org/packages/Microsoft.Data.Sqlite/) package that relies on the `e_sqlite3` native library. This library needs to be accessible and discovered by the scripting runtime and we do that on Windows using `LoadLibrary`.

It is sort of a hack, but it seems to make the scripting runtime aware of the DLL. Preferably we would somehow set the `NATIVE_DLL_SEARCH_DIRECTORIES` property of the AppDomain, but we can't do that once the CoreHost is up and running. We simply come in to late. 

The `LoadLibrary` method can only be used on Windows, but the *nix counterpart, `dlopen` does not do the trick here.

### Resolving compilation dependencies (OmniSharp)

Compilation dependencies (or compile-time dependencies) are meant to be used by `OmniSharp.Script` to provide intellisense for script files.

```c#
LogFactory logFactory = type =>
{
    // Create actual logger here
    return ((level, message) =>
    {
        // Forward log message to actual logger here.
    }); 

bool enableScriptNuGetReferences = true;
var resolver = new CompilationDependencyResolver(logFactory)
var compilationDependencies = resolver.GetDepedencies(targetDirectory, enableScriptNugetReferences);
```

> The reason that we deal with runtime dependencies on the execution side, is that the Roslyn scripting engine currently only supports compiling against runtime assemblies. 
>
> Take a look [here](https://github.com/dotnet/roslyn/issues/6094) for more information.

### Say goodbye to project.json

For a long time during the beta days of .Net Core, it looked like Microsoft was heading for a new project format for .Net projects. `project.json` was introduced as the new lightweight project format and a lot of time and effort was put in to convert projects from the old `csproj` format to this new shiny `project.json` format. After a while or more specifically towards the end of the beta period, Microsoft decided to ditch project.json and go back to csproj. Not the old and rather verbose csproj, but a new simpler csproj format.

`project.json` practically died over night.

OmniSharp.Script still uses `project.json` as a way to provide a "runtime context" from which the compilation dependencies can be retrieved. This `project.json` file can contain references to NuGet packages and hence OmniSharp can provide intellisense for them as well. 

The problem with `project.json` today is that we barely have any tooling left that can understand the format. NuGet stopped supporting `project.json` after version 3.5 and the dotnet sdk never understood the format after going out of beta. So for instance, if you want to use `project.json` today , you will need an outdated [preview](https://github.com/dotnet/core/releases/tag/v1.1.0-preview1) version of the dotnet sdk that most developers do not install. 

IMHO, it is about time that `project.json` dies, also in the context of scripting


seesharper:
The Travis build currently breaks because of this
https://github.com/travis-ci/travis-ci/issues/8552

src/Dotnet.Script.Core/ScriptCompiler.cs
```diff
@@ -146,17 +134,19 @@ public virtual ScriptOptions CreateScriptOptions(ScriptContext context)
...
-            {
-                if (runtimeDependency.AssemblyName.Version != assemblyName.Version)
-                {
-                    _logger.Verbose(
-                        $"Unresolved assembly {assemblyName}. Loading from resolved runtime dependencies at path: {runtimeDependency.Path}");
-                    return loadContext.LoadFromAssemblyPath(runtimeDependency.Path);
-                }
-            }
+        {
+            //var runtimeDependency = runtimeDependencies.SingleOrDefault(r => AssemblyLoadContext.GetAssemblyName(r.Path).Name == assemblyName.Name);
```
filipw: what happened here? 🙂

src/Dotnet.Script.DependencyModel.Nuget/Dotnet.Script.DependencyModel.NuGet.csproj
```diff
@@ -0,0 +1,21 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFrameworks>netstandard2.0;net46</TargetFrameworks>
```
filipw: net46 is for omnisharp, right?

src/Dotnet.Script.DependencyModel/ProjectSystem/csproj.template
```diff
@@ -0,0 +1,8 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
+    <TargetFramework>netcoreapp2.0</TargetFramework>  
+  </PropertyGroup>
+  <ItemGroup>
```
filipw: this empty group can be removed

src/Dotnet.Script/Program.cs
```diff
@@ -142,7 +145,19 @@ private static Task RunCode(string code, string config, bool debugMode, IEnumera
         private static Task Run(bool debugMode, ScriptContext context)
         {
             var logger = new ScriptLogger(Console.Error, debugMode);
-            var compiler = new ScriptCompiler(logger, new ScriptProjectProvider(new ScriptParser(logger), logger));
+            var runtimeDependencyResolver = new RuntimeDependencyResolver(type => ((level, message) =>
```
filipw: I think it's probably time to get rid of `ScriptLogger` completely and just use the MS logger library. this is not related to this PR though, so we can do it later

src/Dotnet.Script.DependencyModel/Context/ScriptDependencyInfoProvider.cs
```diff
@@ -0,0 +1,76 @@
...
+            var context =  ReadDependencyContext(pathToProjectFile);
+            var nugetPackageFolders = GetNuGetPackageFolders(pathToProjectFile);
+            return new ScriptDependencyInfo(context, nugetPackageFolders);
+        }
+
+        private void Restore(string pathToProjectFile)
+        {
+            foreach (var restorer in _restorers)
+            {
+                if (restorer.CanRestore)
```
filipw: it looks like in most cases, this will run both restorers, is this intended?

filipw: oh sorry, I meant it rather will only run the first restorer

src/Dotnet.Script.DependencyModel/Context/ScriptDependencyInfoProvider.cs
```diff
@@ -0,0 +1,76 @@
...
+                // https://github.com/dotnet/core-setup/blob/master/src/managed/Microsoft.Extensions.DependencyModel/DependencyContextJsonReader.cs
+                using (var contextReader = new DependencyContextJsonReader())
+                {
+                    return contextReader.Read(fs);
+                }
+            }
+        }
+
+        private string[] GetNuGetPackageFolders(string pathToProjectFile)
+        {
```
filipw: i think this method may throw in various situations, we should probably try to gracefully exit from here in those cases?

src/Dotnet.Script.DependencyModel/ProjectSystem/ScriptProjectProvider.cs
```diff
@@ -0,0 +1,74 @@
...
+        {
+            var tempDirectory = Path.GetTempPath();
+            var pathRoot = Path.GetPathRoot(targetDirectory);
+            var targetDirectoryWithoutRoot = targetDirectory.Substring(pathRoot.Length);
+            if (pathRoot.Length > 0 && RuntimeHelper.IsWindows())
+            {
+                var driveLetter = pathRoot.Substring(0, 1);
+                targetDirectoryWithoutRoot = Path.Combine(driveLetter, targetDirectoryWithoutRoot);
+            }
+            var pathToProjectJsonDirectory = Path.Combine(tempDirectory, "scripts", targetDirectoryWithoutRoot);
```
filipw: the old name needs an update 🙂

src/Dotnet.Script.DependencyModel/ProjectSystem/ScriptProjectProvider.cs
```diff
@@ -0,0 +1,74 @@
...
+
+            if (!Directory.Exists(pathToProjectJsonDirectory))
+            {
+                Directory.CreateDirectory(pathToProjectJsonDirectory);
+            }
+
+            var pathToProjectJson = Path.Combine(pathToProjectJsonDirectory, "script.csproj");
```
filipw: and here

.travis.yml
```diff
@@ -1,13 +1,13 @@
...
+      osx_image: xcode9      
+      sudo: required      
+      dotnet: 2.0.0    
+before_install:
+  - brew update
 install:
-  - cd build
-  - ulimit -n4096
-  - ./install-dotnet-script.sh  
+  - ulimit -n4096
```
filipw: this is missing space, shouldn't it be `ulimit -n 4096`?

filipw:
beautiful! thank you 🍻🍻🍻
I left a few comments but it's mostly just silly stuff

Let's get this in ASAP

src/Dotnet.Script.DependencyModel/Context/DotnetRestorer.cs
```diff
@@ -0,0 +1,26 @@
...
+        public DotnetRestorer(CommandRunner commandRunner, LogFactory logFactory)
+        {
+            _commandRunner = commandRunner;
+            _logger = logFactory.CreateLogger<DotnetRestorer>();
+        }
+
+        public void Restore(string pathToProjectFile)
+        {
+            _logger.Debug($"Restoring {pathToProjectFile} using the dotnet cli.");            
+            _commandRunner.Execute("dotnet", $"restore {pathToProjectFile}");
```
filipw: no need to pass in the runtime ID here? also, I think we should quote this path like mentioned in the PR here https://github.com/filipw/dotnet-script/pull/139/files

.travis.yml
seesharper: Has always been `-n4096` since we started building on the Mac. A quick test shows that both works.

src/Dotnet.Script.DependencyModel.Nuget/Dotnet.Script.DependencyModel.NuGet.csproj
seesharper: Exactly. The reason is that we must be able to reference this from OmniSharp

src/Dotnet.Script.DependencyModel/Context/ScriptDependencyInfoProvider.cs
seesharper: That is intentional. The 99% case here would be the `DotnetRestorer`. The `NuGetRestorer` is sort of a fallback in the case where a script containing NuGet references is opened in OmniSharp without the `dotnet sdk` installed. In real life this would be a rare scenario

src/Dotnet.Script.DependencyModel/ProjectSystem/ScriptProjectProvider.cs
seesharper: Good catch :) I'll fix :)

src/Dotnet.Script.DependencyModel/ProjectSystem/ScriptProjectProvider.cs
seesharper: 👍

src/Dotnet.Script/Program.cs
seesharper: Agreed. `Dotnet.Script` should use a logging framework such as Microsoft.Extensions.Logging. The reason for keeping the actual logging implementation out of `Dotnet.Script.DependencyModel` is so that we don't run into trouble with different versions of the logging framework.

src/Dotnet.Script.Core/ScriptCompiler.cs
seesharper: Dead code from the past. Removing :)

src/Dotnet.Script.DependencyModel/ProjectSystem/csproj.template
seesharper: The `<ItemGroup>`  is a placeholder for package references. Could have been added from code, but I see no actual harm having it in the template.

src/Dotnet.Script.DependencyModel/Context/DotnetRestorer.cs
seesharper: Fixing quotes. Not passing the runtime id actually cuts the project.assets.json file size in half and still contains all the information we need to resolve runtime dependencies. If we run with a runtime id, we get presented with ALL the native/unmanaged dll's from the core framework and we used to load all of the using `LoadLibrary` which is not needed at all.

filipw: ah good point!

src/Dotnet.Script.DependencyModel/Context/ScriptDependencyInfoProvider.cs
seesharper: Yes, it could throw although something very special must happen in order for it to fail.
I've thought long and hard about what to do with exceptions inside this library. Should we log and eat exceptions or should be just let it bubble up the caller and put the burden of try/catching/log on them? The callers here would be `dotnet.script` and `Omnisharp.Script`. I did not want to do log/rethrow since that would probably cause a duplicate log message, but we could add an exception handler in the `GetDepenencies` method in the dependency resolvers. This handler logs the exception and simply yields an empty list of dependencies. It would save the caller some work.

filipw: yeah sounds like a good idea. let's leave this for now, I'd like to merge it in, and we do that in the future

filipw:
I think I managed to introduce a conflict by merging the other PR, can you please resolve (in this case it would be just override with yours, since dependency resolution changes completely anyway)

seesharper:
No problem. I'll fix it

filipw:
🦄

PR #152, opened 2017-10-17
Updated to Roslyn 2.4.0

filipw:
Also updated xunit


filipw:
yeah OmniSharp will receive the Roslyn update too

Issue #157, opened 2017-10-30
Support local NuGet.config 


seesharper:
We should support having a local `NuGet.Config` file that lets the user specify package sources without modifying the global `NuGet.Config` file.

This should be fairly easy to do. We simply need to copy this file to the same temp folder that contains the `csproj` file

PR #159, opened 2017-10-31
Support for local NuGet.Config

seesharper:
This PR adds support for providing a local `NuGet.Config` file that lets the user specify NuGet configuration without modifying the global `NuGet.Config` file

See #157 for details


filipw:
thanks!

Issue #158, opened 2017-10-30
Propagate the return value from a script as the process exit code


adamralph:
With csi.exe, the `int` returned from a script (it is only allowed to be an `int`) is propagated as the process exit code. E.g. if I return `0` from my script, csi.exe exits with `0` which indicates no error. If I return `1` from my script, `csi.exe` exits with `1`, which indicates an error.

`dotnet-script` should match this behaviour from `csi.exe`. Specifically, we need the runner to behave in this manner to support this proposal for simple-targets-csx: https://github.com/adamralph/simple-targets-csx/issues/49

seesharper:
Makes sense to me :+1 Is there a specific code for compilation errors that we should make sure to match?

adamralph:
This isn't for compilation errors. This is for when the script compiles, and runs to completion with no exception being thrown. At the end of the script an integer is returned, and that integer is propagated to the OS as the exit code of dotnet-script.

adamralph:
E.g.

```C#
// foo.csx

// do some stuff

if (everythingWorkedOK)
{
    return 0;
}
else
{
    return 1;
}
```

filipw:
For compilation errors CSI returns 1.

seesharper:
> This isn't for compilation errors.

I got that :) Was just making sure that we align the exit code for compilation errors as well.

adamralph:
OK 😉 

Bear in mind that the 1 returned by csi.exe could be a compilation error, or it could be a 1 returned from a script, so it doesn't really give you any information. Sure, you could attempt to parse stderr and sniff out the compilation error from there, but then perhaps the script itself is using the roslyn libs. Very difficult to be deterministic there.

filipw:
Closed by #160

PR #160, opened 2017-10-31
Feature/propagate returnvalue

seesharper:
This PR adds support for propagating the return value from scripts.

We also align with CSI to return exit code 1 when a script throws or does not compile.

#158


src/Dotnet.Script.Tests/ScriptExecutionTests.cs
```diff
@@ -75,6 +82,15 @@ public static void ShouldNotPassUnEscapedKnownArgumentToScript()
             Assert.DoesNotContain("-v", result.output);            
         }
 
+        [Fact]
+        public static void ShouldPropagateReturnValue()
+        {
+            var result = Execute($"{Path.Combine("ReturnValue", "ReturnValue.csx")}");
```
filipw: since the errors are 1, let's use a different return value for this "happy path" test 😊

filipw:
looks good!

adamralph:
I'm not sure it's something you also want to align as well, but csi.exe does not let you return anything but an `int` from a script.

filipw:
Yeah that's a good idea. Let's create the script explicitly as `int` here https://github.com/filipw/dotnet-script/pull/160/files#diff-5934cd0deb2fb20e20ed07e1c5eb9e23R155
Any other return type doesn't matter anyway as it's not going to be surfaced.

In CSI using a non-int return creates a compilation error.

src/Dotnet.Script.Tests/ScriptExecutionTests.cs
seesharper: I agree, lets use 42 :)

seesharper:
Are we good now, @filipw ?

filipw:
thanks!

Issue #71, opened 2017-03-13
Add REPL


filipw:
Should be simple enough with current state of Roslyn APIs.

PR #168, opened 2017-11-07
Show interactive errors in red

filipw:
This PR also contains 3 commits I forgot to push into #71


filipw:
looks like I managed to break script execution - the tests ended up starting a REPL instead 😂

PR #170, opened 2017-11-07
uses fully qualified name for Expand-Archive

marking:
uses Microsoft.PowerShell.Archive\Expand-Archive to resolve a problem of running
the install script on a user's machine that has PSCX installed, which also exposes
Expand-Archive


filipw:
thanks!

marking:
no problem. glad to help

PR #173, opened 2017-11-08
removed configuration argument

### LINKED ISSUE #19: Remove "Configuration" argument
It's quite confusing and is only used for `project.json` exporter, we can just infer it from `debug` switch I think.

filipw:
It's not used for anything. 
Fixes #19


filipw:
The only place where it *could* be useful is if we ever start reading more things from `csproj` and they are conditionally available, but I very much doubt that. Let's not pollute the app now.

PR #176, opened 2017-11-16
make sure we install machine wide

andmos:
When rolling out new servers with Ansible or similar provisioning software we often need to be machine wide for all users to be able to use dotnet-script.


seesharper:
Thanks

seesharper:
@filipw Took the liberty of merging this :)

PR #178, opened 2017-11-19
Force LF EOL for sh scripts

filipw:
I just copied the whole file from Omnisharp repo 😄


seesharper:
The branch is out of date. I'll fix it

PR #179, opened 2017-11-20
Feature/script packages

seesharper:
This PR adds initial and basic support for loading script packages using the `#load nuget:packagename, 1.2.3` directive.

Changes has been made both to the `RuntimeDependencyResolver` and the `CompilationDependencyResolver` to accommodate for script files resolved from NuGet packages. This means breaking changes for the `Dotnet.Script.DependencyModel`  and the `Dotnet.Script.DependencyModel.NuGet` package. These package has been bumped to 1.0.0 to reflect this.

We create and pack all test packages automatically from within the test project and we can reproduce just about any package scenario which is good for the road ahead.

Everything is prepared also for the changes needed in `OmniSharp.Script` so once we merge and publish these packages, we can PR `OmniSharp.Script`with changes. The changes needed in `OmniSharp.Script` are pretty minor.

This is a pretty big PR, but this represents the MVP of script packages and we are in a much better spot for further improvements on the dependency model. 

The documentation has been updated and included here for reference.

### Script Packages

Script packages are a way of organizing reusable scripts into NuGet packages that can be consumed by other scripts. This means that we now can leverage scripting infrastructure without the need for any kind of bootstrapping. 

#### Creating a script package

A script package is just a regular NuGet package that contains script files inside the `content` or `contentFiles` folder.

The following example shows how the scripts are laid out inside the NuGet package according to the [standard convention](https://docs.microsoft.com/en-us/nuget/schema/nuspec#including-content-files) .

```shell
└── contentFiles
    └── csx
        └── netstandard2.0
            └── main.csx
```

This example contains just the `main.csx` file in the root folder, but packages may have multiple script files either in the root folder or in subfolders below the root folder. 

When loading a script package we will look for an entry point script to be loaded. This entry point script is identified by one of the following.

- A script called `main.csx` in the root folder  
- A single script file in the root folder

If the entry point script cannot be determined, we will simply load all the scripts files in the package.

> The advantage with using an entry point script is that we can control loading other scripts from the package. 

#### Consuming a script package

To consume a script package all we need to do specify the NuGet package in the `#load `directive.

The following example loads the [simple-targets](https://www.nuget.org/packages/simple-targets-csx) package that contains script files to be included in our script.

```c#
#! "netcoreapp2.0"
#load "nuget:simple-targets-csx, 6.0.0"

using static SimpleTargets;
var targets = new TargetDictionary();

targets.Add("default", () => Console.WriteLine("Hello, world!"));

Run(Args, targets);
```

Note: Debugging also works for script packages so that we can easily step into the scripts that are brought in using


src/Dotnet.Script.Core/ScriptCompiler.cs
```diff
@@ -15,11 +16,17 @@
...
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+using RuntimeAssembly = Dotnet.Script.DependencyModel.Runtime.RuntimeAssembly;
 
 namespace Dotnet.Script.Core
 {
     public class ScriptCompiler
     {
+        // Note: Windows only, Mac and Linux needs something else?
+        [DllImport("Kernel32.dll")]
+        private static extern IntPtr LoadLibrary(string path);
```
filipw: moving this here, doesn't this affect OmniSharp?

src/Dotnet.Script.DependencyModel/ScriptPackage/ScriptFilesDependencyResolver.cs
```diff
@@ -0,0 +1,138 @@
...
+        public ScriptFilesDependencyResolver(LogFactory logFactory)
+        {            
+            _logger = logFactory.CreateLogger<ScriptFilesDependencyResolver>();
+        }
+
+        public string[] GetScriptFileDependencies(string path, string[] nugetPackageFolders)
+        {
+            var result = new List<string>();
+            var packagePath = GetPackageFullPath(path, nugetPackageFolders);
+            var allScriptFiles = Directory.GetFiles(packagePath, "*.csx", SearchOption.AllDirectories);
```
filipw: should we perhaps discover script files only once, somewhere in the beginning and pass that list to both `ScriptFilesDependencyResolver` and `ScriptProjectProvider`?

filipw: oh wait, this happens per nuget folder, ok

src/Dotnet.Script.DependencyModel/Runtime/RuntimeDependencyResolver.cs
```diff
@@ -67,21 +66,32 @@ private IEnumerable<RuntimeDependency> GetDependenciesInternal(string pathToProj
             List<string> nuGetPackageFolders = dependencyInfo.NugetPackageFolders.ToList();
             nuGetPackageFolders.Add(RuntimeHelper.GetPathToNuGetStoreFolder());
 
-            var runtimeDepedencies = new HashSet<RuntimeDependency>();
+            var runtimeDepedencies = new List<RuntimeDependency>();
```
filipw: typo

src/Dotnet.Script.DependencyModel.Nuget/NuGetSourceReferenceResolver.cs
```diff
@@ -0,0 +1,92 @@
...
+namespace Dotnet.Script.DependencyModel.NuGet
+{
+    /// <summary>
+    /// A <see cref="SourceReferenceResolver"/> decorator that handles
+    /// references to NuGet packages in scripts.  
+    /// </summary>
+    public class NuGetSourceReferenceResolver : SourceReferenceResolver
+    {
+        private readonly SourceReferenceResolver _sourceReferenceResolver;
+        private readonly IDictionary<string, IReadOnlyList<string>> _scriptMap;
```
filipw: since this is part of a public API I think for easier comprehension we should wrap it into a class rather than a fairly cryptic `IDictionary<string, IReadOnlyList<string>>`

src/Dotnet.Script.DependencyModel.Nuget/NuGetSourceReferenceResolver.cs
```diff
@@ -0,0 +1,92 @@
...
+using System.Text.RegularExpressions;
+using Microsoft.CodeAnalysis;
+
+namespace Dotnet.Script.DependencyModel.NuGet
+{
+    /// <summary>
+    /// A <see cref="SourceReferenceResolver"/> decorator that handles
+    /// references to NuGet packages in scripts.  
+    /// </summary>
+    public class NuGetSourceReferenceResolver : SourceReferenceResolver
```
filipw: in order to make this work with the REPL there needs to be some sort of a public method that allows us to mutate the script map. Something like `WithScriptMap()` or so

So when I start a REPL, original options are created without script map. But as soon as I do `#load` from nuget, the resolver should get a new map discovered based on the current `#load`(or many `#loads`). I'm just not sure if we do it multiple times in a REPL session, should we keep appending to the map or replace it every time, I think replace it every time since the earlier compilations will already have been completed

src/Dotnet.Script.Core/Interactive/InteractiveRunner.cs
```diff
@@ -70,7 +70,7 @@ public InteractiveRunner(ScriptCompiler scriptCompiler, ScriptLogger logger, Scr
                 }
                 else
                 {
-                    var lineDependencies = ScriptCompiler.RuntimeDependencyResolver.GetDependenciesFromCode(CurrentDirectory, input);
+                    var lineDependencies = ScriptCompiler.RuntimeDependencyResolver.GetDependenciesFromCode(CurrentDirectory, input).SelectMany(rtd => rtd.Assemblies).Distinct();
```
filipw: this hypothetical `WithScriptMap` against the resolver inside `ScriptOptions` should be invoked from here

src/Dotnet.Script.Core/ScriptCompiler.cs
```diff
@@ -68,11 +75,11 @@ public ScriptCompiler(ScriptLogger logger, RuntimeDependencyResolver runtimeDepe
             RuntimeDependencyResolver = runtimeDependencyResolver;
         }
 
-        public virtual ScriptOptions CreateScriptOptions(ScriptContext context)
-        {
+        private ScriptOptions CreateScriptOptions(ScriptContext context, IDictionary<string, IReadOnlyList<string>> scriptMap)
```
filipw: I would rather keep it `public virtual` I have some ideas on how to consume this core library i.e. inside an ASP.NET Core app and would like to keep this overrideable. I'd suggest to pass entire `IList<RuntimeDependency>` to this method and pick up script map inside it, this way the public API is cleaner

src/Dotnet.Script.DependencyModel.Nuget/NuGetSourceReferenceResolver.cs
seesharper: I agree. The reason I did that here was to keep the number of dependencies on a low down for the `Dotnet.Script.DependencyModel.NuGet` package. Since we bring back a reference to this assembly as a hack inside the `NuGetMetadataReferenceResolver`. If we reference `Dotnet.Script.DependencyModel` from this package, I am afraid that we bring in types from that assembly too. Not sure if that is an issue though. I'll take a look at it

src/Dotnet.Script.DependencyModel/Runtime/RuntimeDependency.cs
```diff
@@ -1,26 +1,27 @@
-﻿using System.Reflection;
+﻿using System.Collections.Generic;
+using System.Reflection;
 
 namespace Dotnet.Script.DependencyModel.Runtime
 {
-    public class RuntimeDependency
+    public class RuntimeAssembly
```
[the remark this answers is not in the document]
seesharper: Of course :)

src/Dotnet.Script.DependencyModel.Nuget/NuGetSourceReferenceResolver.cs
```diff
@@ -0,0 +1,92 @@
...
+
+            return _sourceReferenceResolver.NormalizePath(path, baseFilePath);
+        }
+
+        public override string ResolveReference(string path, string baseFilePath)
+        {
+            if (path.StartsWith("nuget", StringComparison.OrdinalIgnoreCase))
+            {
+                var packageName = PackageNameMatcher.Match(path).Groups[1].Value;
+                var scripts = _scriptMap[packageName];
```
filipw: this will unfortunately have the same issue as the nuget resolver for `#r`, in that it will not work with OmniSharp in real time (as you type) but only on initial file opening.

We will need to look at injecting sort of a state machine into these resolver and if resolve the references from the state machine; if not found there, we run `dotnet restore` again. Not very elegant but I don't think there is a better way to do it with the current Roslyn extensibility points

build/Choco.csx
```diff
@@ -1,3 +1,4 @@
+#! "netcoreapp1.1"
```
filipw: is there any reason for this?

build/Build.csx
```diff
@@ -1,12 +1,16 @@
...
 #load "Choco.csx"
+#load "NuGet.csx"
+#load "FileUtils.csx"
 
 using System.Runtime.InteropServices;
 
-var rootArg = Args.FirstOrDefault() ?? "..";
-var root = Path.GetFullPath(rootArg);
+
+var currentFolder = Path.GetDirectoryName(GetScriptPath());
```
filipw: 👍

src/Dotnet.Script.Core/ScriptCompiler.cs
seesharper: No. Omnisharp does not deal with native assets. Those are only needed at runtime. I've simply moved it out of the `RuntimeDependencyResolver` since that class should be about bringing back the runtime dependencies and not about actually loading them. That we do in the `ScriptCompiler` class.

filipw:
thanks! looks awesome - massive work! I left a few stupid comments here and there 🙂

src/Dotnet.Script.DependencyModel.Nuget/NuGetSourceReferenceResolver.cs
filipw: maybe let's leave it as is for now. once we have an idea of how to address https://github.com/filipw/dotnet-script/pull/179#discussion_r152201758 this should not be needed anymore, as the map will not be passed into the `NuGetSourceReferenceResolver` anymore, it will build itself up from within

src/Dotnet.Script.DependencyModel.Nuget/Dotnet.Script.DependencyModel.NuGet.csproj
```diff
@@ -8,7 +8,7 @@
     <RepositoryUrl>https://github.com/filipw/dotnet-script.git</RepositoryUrl>
     <RepositoryType>git</RepositoryType>
     <PackageTags>script;csx;csharp;roslyn;nuget</PackageTags>
-    <Version>0.2.0</Version>
+    <Version>1.0.0</Version>
```
filipw: the version can be `0.3.0`. Before `1.0` it's not a violation of semver to have breaking changes (that's why scriptcs stayed on pre 1.0 so long 😀 )

src/Dotnet.Script.Core/ScriptCompiler.cs
seesharper: No problem. Fixing

src/Dotnet.Script.DependencyModel.Nuget/Dotnet.Script.DependencyModel.NuGet.csproj
seesharper: Actually, I did not know you could do that pre 1.0.0. :) Changing to 0.3.0 for both `DependencyModel` packages.

build/Choco.csx
seesharper: No reason. Removing 😸

src/Dotnet.Script.DependencyModel/Runtime/RuntimeDependencyResolver.cs
seesharper: Should think I could spell "dependencies" by now 😄 Fixing

src/Dotnet.Script.DependencyModel.Nuget/NuGetSourceReferenceResolver.cs
seesharper: I'll look into it. Will create a test for this in the `InteractiveRunnerTests` class

filipw: actually `ScriptOptions` itself is immutable so it needs to be reset every time. This means that it should be enough to just do `_scriptOptions = _scriptOptions.WithSourceResolver(...)` inside the `Interactiverunner` and pass in a whole new instance of `NuGetSourceReferenceResolver` with a new source map extracted from the current REPL submission. I'm just not sure if we can safely reset the resolver like that, we should be able though, as the previous REPL submissions are already compiled in memory

filipw: but again, this will be not needed anymore as soon as we have a good story for "self-contained" resolvers (such that don't need to modify `ScriptOptions`) which I really hope we can have at some point

src/Dotnet.Script.Core/Interactive/InteractiveRunner.cs
```diff
@@ -70,8 +72,14 @@ public InteractiveRunner(ScriptCompiler scriptCompiler, ScriptLogger logger, Scr
                 }
                 else
                 {
-                    var lineDependencies = ScriptCompiler.RuntimeDependencyResolver.GetDependenciesFromCode(CurrentDirectory, input);
-
+                    var lineRuntimeDependencies =
+                        ScriptCompiler.RuntimeDependencyResolver.GetDependenciesFromCode(CurrentDirectory, input).ToArray();
+                    var lineDependencies = lineRuntimeDependencies.SelectMany(rtd => rtd.Assemblies).Distinct();
+
+                    var scriptMap = lineRuntimeDependencies.ToDictionary(rdt => rdt.Name, rdt => rdt.Scripts);
```
filipw: did this work? btw, we should only reset the resolver when the `scriptMap` is not empty

seesharper: It did work, although I think the `InteractiveRunner` needs some more work.  I noticed that it takes another approach the first time it is executed, `RunFirstScript` . I did not want to mess with that too much since I guess there is a good reason for that special handling 😄

seesharper: btw, I also added a test, `ScriptPackageReference` in the `InteractiveRunnerTests` class for this

filipw: yeah you are right, there is a good reason for this, but this is also a bug 😀 We can fix this in a separate PR

seesharper: Awesome 👍

filipw:
@seesharper this should be good to merge now?

seesharper:
Yes, we are good 👍

filipw:
🎉

seesharper:
I'll create the PR for `OmniSharp.Script`, once we have released 0.16.0 along with the new `DependencyModel` packages. You want to apply some REPL 💓  before we release?

PR #184, opened 2017-11-24
Better errors

filipw:
- we don't drop a full stacktrace in REPL anymore, instead we use the same formatter as CSI to provide a nice single-line message
 - errors in REPL, as well as compilation+runtime errors in scripts are printed into the console in red color
 - if a runtime error happens in a script, the runner exit code is set to the `HResult` of the exception (this is a recent change in CSI which will likely go out in 2.6.0). COmpilation errors still result in exit code 1

### DIFF
--- a/src/Dotnet.Script.Core/Interactive/InteractiveRunner.cs
+++ b/src/Dotnet.Script.Core/Interactive/InteractiveRunner.cs
@@ -103,6 +103,13 @@ namespace Dotnet.Script.Core
                     _globals.Print(_scriptState.ReturnValue);
                 }
             }
+            catch (CompilationErrorException e)
+            {
+                foreach (var diagnostic in e.Diagnostics)
+                {
+                    Console.WritePrettyError(diagnostic.ToString());
+                }
+            }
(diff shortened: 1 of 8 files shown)


filipw:
Ah one more thing, the runner runtime exceptions still print the full stacktrace. There is a future change in Roslyn that will allows us to trim this down, but for now it stays as it is. Full stack trace is valuable of course cause it has line error numbers

filipw:
not sure why but `HResult` for an exception is different on Windows and nix??

seesharper:
I have no idea. Could be somehow make the Assert less specific ?

filipw:
> I have no idea. Could we somehow make the Assert less specific ?

OK turns out the idea is nonsense since the exit code range is different on nix
https://github.com/dotnet/cli/issues/918#issuecomment-173338483.

I've reverted this change, and it will have to be reverted from CSI. 
I'll propose a unified exit code, different than compilation exit code (i.e. 2) for all runtime exceptions there, then we can align.

seesharper:
Sound like a good idea

PR #185, opened 2017-11-24
Fixes #183

### LINKED ISSUE #183: Document using dotnet-script with piping
Came across a pretty interesting use case for `dotnet-script` today. I was asked by @andmos if we could pipe something into `dotnet-script` 

Something like 
```
echo "This is some text" | dotnet script UpperCase.csx 
```

This is what we came up with (`UpperCase.csx`)

```
#! "netcoreapp2.0"
#r "nuget: NetStandard.Library, 2.0.0"

using (var streamReader = new StreamReader(Console.OpenStandardInput()))
{    
    Write(streamReader.ReadToEnd().ToUpper()); 
}
```

Then we can do 

```
$ echo "This is some text" | dotnet script UpperCase.csx
THIS IS SOME TEXT
```

@filipw Although this is not specific to `dotnet-script` , it might be a very useful scenario to mention in the documentation. What do you think? Pretty nice, huh? 😄

seesharper:
This PR adds documentation for how to use `dotnet-script`with piping. Also added a section how to debug these type of scripts.

### DIFF
--- a/README.md
+++ b/README.md
@@ -194,6 +194,89 @@ Run(Args, targets);
 
  
 
+### Piping 
+
+The following example shows how we can pipe data in and out of a script.
+
+The `UpperCase.csx` script simply converts the standard input to upper case and writes it back out to standard output.
+
+```csharp
+#! "netcoreapp2.0"
+#r "nuget: NetStandard.Library, 2.0.0"
+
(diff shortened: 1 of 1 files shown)


seesharper:
@filipw Are we good to merge this?

filipw:
thanks!

PR #186, opened 2017-11-24
Added Microsoft.DotNet.PlatformAbstractions to Build.csx

seesharper:
This PR adds the the Microsoft.DotNet.PlatformAbstractions NuGet package to the `Build.csx` file. This is needed from now on since we don't get this through the inherited assemblies anymore.  
Does not change anything, it just prepares us for when we start building with 0.16.0.

Last PR from me before 0.16.0 😄


filipw:
thanks

Issue #200, opened 2017-12-06
Let `install.sh` support selection of version to install 


andmos:
When provisioning `dotnet-script` to serverfarms via Chef or Ansible, it should be able to keep some state of what is installed and select a specific version. With Chocolatey this is easy, since both Chef and Ansible has modules that let you select version or always the latest version of the package. Will take a look with @seesharper on this!

SlowLogicBoy:
smething like svm for scriptcs ?

andmos:
May be a bit overkill for my usecase right now, but tools like `svm` are really clever and should be looked at not-so-long down the road!

filipw:
we could repurpose svm 😀

andmos:
The `svm` approach is great for local development environments and can also be used for controlling servers. The reason I call it overkill just for _me_ is the way I use Ansible to provision out software on the farm. For `dotnet-script` on Windows i do:  
```yml
- name: Install dotnet.script
   win_chocolatey: name=dotnet.script  state=latest source="{{ ChocoFeedUrl }}"
```  
So what basically happens here is that the Chocolatey module keeps track on the installed version and returns `OK` if I have the latest package, but gives me an `CHANGED` event if it is installing a new one. 
I can also choose to use a `version=` variable here. So on Windows, this is great. Full control of versions, while keeping track of the installed version. 

For installing on Linux i grab the `install.sh` script and run it directly. To check if it is installed I check some files. If I choose to run `install.sh` each time, I will always get `CHANGED` event even if it didn't install anything new, making the state check a bit useless. It would be nice to have the same option of control here too. 

The `svm` approach will solve it, or just some "smarter" checks in the `install.sh` script. We can figure out something neat 👍

seesharper:
Covered by #201

andmos:
If anybody comes across this post later, my Ansible role became: 

```yml
- name: Run installDotNetScript.sh
  shell: ./install.sh
  args:
    chdir: /tmp/
  register: installDotNetScript
  changed_when: '"already installed" not in installDotNetScript.stdout'
```

PR #201, opened 2017-12-06
Added support specifying version when installing on *nix

seesharper:
This PR adds support for specifying the version to install on *nix (`install.sh`).

We also skip installation if the target version is already installed.

The default has not changed which is to install the latest version. 

See #200 for more details

Credits to @andmos for helping out with the bash stuff


andmos:
LGTM

seesharper:
@filipw Are we good to merge this? 😄

PR #202, opened 2017-12-06
Added support for binding redirect and highest version selection

### LINKED ISSUE #198: Assembly Binding?
```csharp
#! "netcoreapp2.0"
#r "nuget:NuGet.Client,*"
#r "nuget:NuGet.Configuration,*"
//#r "nuget:Nuget.Common,*"

using System.Linq;
using NuGet;
using NuGet.Configuration;
using NuGet.Repositories;

var repo = new NuGetv3LocalRepository(SettingsUtility.GetGlobalPackagesFolder(Settings.LoadDefaultSettings(null)));
var packages = repo.FindPackagesById("NuGet.Client");
var pkg = packages.Last();
var nuspec = pkg.Nuspec;
```

I get: `System.IO.FileNotFoundException: Could not load file or assembly 'NuGet.Common, Version=4.2.0.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35'. The system cannot find the file specified.
File name: 'NuGet.Common, Version=4.2.0.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35' ---> System.IO.FileNotFoundException: Could not load the specified file.`

I've tried specifying versions, etc. Can't figure out how to make this work :).

I don't know if it's an Assembly Binding issue or NuGet.Client has something nasty and it's looking for that dll in working directory.

P.S. this works in .csproj

seesharper:
This PR adds support for redirecting a request for an assembly to the corresponding assembly from the dependency context. 

Also improved the logic around adding references to runtime assemblies where we now ensure that we load the assembly with the highest version number in the case where the dependency contexts contains multiple versions of the same runtime assembly. 

This PR also fixes #198

### DIFF
--- a/src/Dotnet.Script.Core/ScriptCompiler.cs
+++ b/src/Dotnet.Script.Core/ScriptCompiler.cs
@@ -4,6 +4,7 @@ using System.Collections.Immutable;
 using System.Linq;
 using System.Reflection;
 using System.Runtime.InteropServices;
+using System.Runtime.Loader;
 using Microsoft.CodeAnalysis;
 using Microsoft.CodeAnalysis.CSharp.Scripting;
 using Microsoft.CodeAnalysis.Scripting;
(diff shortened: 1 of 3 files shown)


seesharper:
@filipw At least we came to the same conclusion 😄 Actually this one is so important that we maybe should consider a service release (0.16.1). This PR does not affect OmniSharp so there is no reason to bump the depenencymodel package versions 👍

PR #205, opened 2017-12-07
Bugfix/core assemblies

### LINKED ISSUE #204: Missing core assembly references
@filipw  We have a pretty serious issue here probably because we now don't add inherited assemblies anymore.

Consider the following script 

```
#! "netcoreapp2.0"
#r "nuget:NetStandard.Library, 2.0.1"

using System;
using System.Net;

Console.WriteLine(typeof(WebProxy));
```

The `WebProxy` class comes from the `System.Net.WebProxy.dll` and is part of NetCoreApp2.0. 

When executing the script in debug mode we get 

```
$ dotnet script -d helloworld.csx
Current runtime is 'win'.
Creating project file for *.csx files found in C:\scriptdemo using netcoreapp2.0 as the default framework.
Parsing C:\scriptdemo\helloworld.csx
Project file saved to C:\Users\bri\AppData\Local\Temp\scripts\C\scriptdemo\script.csproj
<?xml version="1.0" encoding="utf-8"?>
  <PropertyGroup>
    <OutputType>Exe</OutputType>
    <TargetFramework>netcoreapp2.0</TargetFramework>
  </PropertyGroup>
  <ItemGroup>
    <PackageReference Include="NetStandard.Library" Version="2.0.1" />
  </ItemGroup>
</Project>
2.1.2
Restoring C:\Users\bri\AppData\Local\Temp\scripts\C\scriptdemo\script.csproj using the dotnet cli.
  Restoring packages for C:\Users\bri\AppData\Local\Temp\scripts\C\scriptdemo\script.csproj...
  Generating MSBuild file C:\Users\bri\AppData\Local\Temp\scripts\C\scriptdemo\obj\script.csproj.nuget.g.props.
  Restore completed in 166,19 ms for C:\Users\bri\AppData\Local\Temp\scripts\C\scriptdemo\script.csproj.
Reading dependency context from C:\Users\bri\AppData\Local\Temp\scripts\C\scriptdemo\script.csproj
Resolved NuGet package folders: C:\Users\bri\.nuget\packages\;C:\Program Files\dotnet\sdk\NuGetFallbackFolder
Adding reference to an inherited dependency => Microsoft.CodeAnalysis
Adding reference to an inherited dependency => Microsoft.CodeAnalysis.CSharp
Adding reference to an inherited dependency => Microsoft.CodeAnalysis.CSharp.Scripting
Adding reference to an inherited dependency => Microsoft.CodeAnalysis.Scripting
C:\scriptdemo\helloworld.csx(7,26): error CS0246: The type or namespace name 'WebProxy' could not be found (are you missing a using directive or an assembly reference?)
```

We can also see from the log that we do not add any runtime assemblies from netcoreapp2.0 so I guess that is why we see this error.

> Note Compilation assemblies for OmniSharp are not affected. They resolve correctly 

We need to be able to get all these assemblies from the dependency context loaded from the csproj file and that means that we probably need to specify the runtime identifier when restoring 

So instead of 

```
dotnet restore script.csproj
```

we need to do 

```
dotnet restore script.csproj -r [runtimeidentifier]  -> e-g win10-x64
```

The `DependencyContext` class has a method called `GetRuntimeAssemblyNames(runtimeIdentifer)` that will return a complete list of runtime assembly names if and only if we restore using the runtime identifier.

We should not go back to pulling these assemblies from the host (inheritedassemblies), but we should fix this in the `RuntimeDependencyResolver` class and make sure that we return the "core" assemblies so that they can be added in the `ScriptCompiler` 

I'm looking into this now

seesharper:
This PR fixes #204 and now ensures that we load all runtime assemblies from NetCoreApp2.0

The template for `dotnet script init` has been updated so that it does not include the NuGet reference to `NetStandard.Library`

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,7 +10,7 @@ before_install:
   - chmod +x dotnet-install.sh
   - ./dotnet-install.sh --channel 2.0
 install:
-  - ulimit -n4096  
+  - ulimit -n8192  
 script:  
   - chmod +x build.sh
   - ./build.sh
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,7 @@
 language: csharp
 dotnet: 2.0.0
+mono:
+ -latest
 matrix:
   include:
     - os: linux
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,17 +1,15 @@
 language: csharp
+mono: none
+dotnet: 2.0.0
 matrix:
   include:
+    - os: linux
+      dist: trusty      
+      sudo: false
     - os: osx
       osx_image: xcode9      
-      sudo: required        
-before_install:
-  - brew update
-  - curl https://dot.net/v1/dotnet-install.sh -o dotnet-install.sh
-  - chmod +x dotnet-install.sh
-  - ./dotnet-install.sh --channel 2.0
-install:
-  - ulimit -n8192  
+      sudo: required               
 script:  
+  - ulimit -n8192  
   - chmod +x build.sh
   - ./build.sh
-  
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,4 @@
 language: csharp
-mono: none
 dotnet: 2.0.0
 matrix:
   include:
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 language: csharp
 dotnet: 2.0.0
 mono:
- -latest
+ - latest
 matrix:
   include:
     - os: linux
--- a/README.md
+++ b/README.md
@@ -59,17 +59,15 @@ docker run -it dotnet-script --version
 Our typical `helloworld.csx` might look like this
 
 ```
-#! "netcoreapp1.1"
-#r "nuget:NetStandard.Library,1.6.1"
-
+#! "netcoreapp2.0"
 Console.WriteLine("Hello world!");
 ```
 
 Let us take a quick look at what is going on here.
 
-`#! "netcoreapp1.1"` tells OmniSharp to resolve metadata in the context of a`netcoreapp1.1` application.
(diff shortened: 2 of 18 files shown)


seesharper:
Okay, that did not turn out so well on MacOS. Looking into it

seesharper:
Just for reference on MacOS 

Ran this for the generated csproj file
```
dotnet publish -r osx.10.13-x64
```

Got 
```
/usr/local/share/dotnet/sdk/NuGetFallbackFolder/microsoft.netcore.app/2.0.0/build/
netcoreapp2.0/Microsoft.NETCore.App.targets(19,5): 
error : Project is targeting runtime 'osx.10.13-x64' but did not resolve any runtime-specific 
packages for the 'Microsoft.NETCore.App' package.  
This runtime may not be supported by .NET Core. 
[/private/var/folders/ws/36c_t7zj4hn5dg8r0nq4whsm0000gn/T/scripts/Users/bernhardrichter
/Github/dotnet-script/src/Dotnet.Script.Tests/TestFixtures/Issue204/script.csproj]
```

No idea what is going on here

seesharper:
Okay, works with RID `osx-x64` 👍

seesharper:
Yeah, too many open files. I've seen this before on MacOS. Probably because xUnit is running in parallel. I'll try `ulimit -n[skyhigh]`

seesharper:
This might be a good time to add Linux to the build matrix on Travis. Just to make sure that RID's are okay there as well 👍

seesharper:
Finally, this took the better part of my day to get right. But we now have builds for both MacOS and Linux in addition to our Windows(AppVeyor) build. The last 7 commits are just minor fixes to casing issues and such.  Fun fact. Commits from 3 different operating systems 😃 
@filipw Are we good to go?

filipw:
wonderful!

PR #211, opened 2017-12-15
Handle #load and #r in eval

### LINKED ISSUE #203: "eval" does not add NuGet packages to the project file
```
dotnet script -d eval '#r \"nuget:AutoMapper, 9.1.0\"; Console.WriteLine(typeof(MapperConfiguration));'
```

We should see a project file containing the package reference here 

```
Current runtime is 'win'.
Creating project file for *.csx files found in C:\repldemo using netcoreapp2.0 as the default framework.
Project file saved to C:\Users\bri\AppData\Local\Temp\scripts\C\repldemo\script.csproj
<?xml version="1.0" encoding="utf-8"?>
  <PropertyGroup>
    <OutputType>Exe</OutputType>
    <TargetFramework>netcoreapp2.0</TargetFramework>
  </PropertyGroup>
  <ItemGroup></ItemGroup>
</Project>
2.1.2
Restoring C:\Users\bri\AppData\Local\Temp\scripts\C\repldemo\script.csproj using the dotnet cli.
  Restore completed in 16,86 ms for C:\Users\bri\AppData\Local\Temp\scripts\C\repldemo\script.csproj.
Reading dependency context from C:\Users\bri\AppData\Local\Temp\scripts\C\repldemo\script.csproj
Resolved NuGet package folders: C:\Users\bri\.nuget\packages\;C:\Program Files\dotnet\sdk\NuGetFallbackFolder
Adding reference to an inherited dependency => Microsoft.CodeAnalysis
Adding reference to an inherited dependency => Microsoft.CodeAnalysis.CSharp
Adding reference to an inherited dependency => Microsoft.CodeAnalysis.CSharp.Scripting
Adding reference to an inherited dependency => Microsoft.CodeAnalysis.Scripting
(1,29): error CS1025: Single-line comment or end-of-line expected
Microsoft.CodeAnalysis.Scripting.CompilationErrorException: Script compilation failed due to one or more errors.
   at Dotnet.Script.Core.ScriptCompiler.CreateCompilationContext[TReturn,THost](ScriptContext context) in C:\Github\fresh\dotnet-script\src\Dotnet.Script.Core\ScriptCompiler.cs:line 151
   at Dotnet.Script.Core.ScriptRunner.Execute[TReturn,THost](ScriptContext context, THost host) in C:\Github\fresh\dotnet-script\src\Dotnet.Script.Core\ScriptRunner.cs:line 32
   at Dotnet.Script.Program.<>c__DisplayClass3_0.<<Wain>b__3>d.MoveNext() in C:\Github\fresh\dotnet-script\src\Dotnet.Script\Program.cs:line 72
```

filipw:
Fixes #203 

Note that at the moment not just the nuget versions, but `#load` and `#r` didn't work in `eval` at all. This PR adds a syntax rewriter which injects an empty line after them to fix the compilation error.


filipw:
Looks like I broke something by accident :)

seesharper:
A good thing we have CI build and unit tests :)

seesharper:
Maybe this just needs to be updated with the base branch

src/Dotnet.Script.DependencyModel/ProjectSystem/ScriptProjectProvider.cs
```diff
@@ -25,7 +26,7 @@ public string CreateProjectForRepl(string code, string targetDirectory, string d
         {
             var parseresult = _scriptParser.ParseFromCode(code);
 
-            targetDirectory = Path.Combine(targetDirectory, "REPL");
+            targetDirectory = Path.Combine(targetDirectory, "interactive", Guid.NewGuid().ToString());
```
seesharper: Would this mean that we get a different temp folder every time it is executed? If so, would the temp folder pile up with csproj files ? 😄

filipw: so each time we need to resolve a a REPL #load/#r or invoke an eval, we should give it a unique context - otherwise we can have collisions (which is why tests were failing). For example, at the moment having two simultaneous REPL sessions from the same folder potentially creates all kinds of problems. The same applies to eval - if you issue 5 simultaneous evals they'd conflict with each other
I agree it's not perfect but I don't have a better idea - unless we say we allow 1 REPL session per folder and 1 concurrent eval per folder. This way we can drop the guid and just use `repl`/`eval` markers. Maybe this would be less invasive for now?

seesharper: I think that would be better. On windows these temp folder are not cleared upon restart and we could easily end up with a ton of these temp folders. On *nix they are cleared, but *nix machines are "never" restarted 😄

seesharper:
LGTM 🎄🍻

PR #212, opened 2017-12-18
Housekeeping/deadcode

seesharper:
This PR removes the `foo.csx` fixture that we don't use anymore in addition to the `launchSettings.json` file that I'm pretty sure we don't need anymore either.


filipw:
thanks definitely not needed

Issue #214, opened 2017-12-23
MissingMethodException using Ravendb.Client and NodaTime.Serialization.JsonNet


germanftorres:
Hi!

I get an exception while referencing various packages that depend on Json.NET (RavenDB.Client and NodaTime.Serialization.JsonNET.  I guess it's related to assembly binding redidrect.  Creating a dotnet new console application with the exact same packages runs fine.

Any ideas what's going on?

Thanks!!

Germán

```csharp
#! "netcoreapp2.0"
#r "nuget: NodaTime, 2.0.0"
#r "nuget: NodaTime.Serialization.JsonNet, 2.0.0"
#r "nuget: RavenDB.Client, 4.0.0-rc-40025"

using System;
using System.Collections.Generic;
using System.Linq;
using NodaTime;
using NodaTime.Serialization.JsonNet;

var converter = NodaConverters.InstantConverter;
Console.WriteLine("Hello World!");

```
```
System.MissingMethodException: Method not found: 'Newtonsoft.Json.JsonConverter NodaTime.Serialization.JsonNet.NodaConverters.get_InstantConverter()'.
   at Submission#0.<<Initialize>>d__0.MoveNext()
   at System.Runtime.CompilerServices.AsyncTaskMethodBuilder`1.Start[TStateMachine](TStateMachine& stateMachine)
   at Submission#0.<Initialize>()
   at Submission#0.<Factory>(Object[] submissionArray)
   at Microsoft.CodeAnalysis.Scripting.ScriptExecutionState.<RunSubmissionsAsync>d__9`1.MoveNext()

```

seesharper:
Hi and thanks for reporting this. We have confirmed this to be a bug and will be looking into it. 

What we have found so far is that RavenDb pulls in Newtonsoft.JSon 10.0.3 which is the version that we add as a reference to the script. The same version (10.0.3) is also loaded if we create a standard console app with these dependencies. The examples here are simplified to pull in Newtonsoft.Json 10.0.3 directly instead of via RavenDB: The error is still present. 

ConsoleApp.csproj
```
<Project Sdk="Microsoft.NET.Sdk">

  <PropertyGroup>
    <OutputType>Exe</OutputType>
    <TargetFramework>netcoreapp2.0</TargetFramework>
  </PropertyGroup>

  <ItemGroup>
    <PackageReference Include="NodaTime" Version="2.2.3" />
[...]

PR #215, opened 2017-12-25
Updated dependencies

seesharper:
This PR updates `Microsoft.Extensions.DependencyModel` from `2.0.3` to `2.0.4` and updates the scripting libraries to `2.6.1`. It also adds an explicit NuGet reference to `Newtonsoft.Json` with version `10.0.3`  which solves #214 because `Microsoft.Extensions.DependencyModel` pulls in version `9.0.1` which is too low in many cases. `Newtonsoft.Json` is one of the few dependencies that we indirectly rely on via `Microsoft.Extensions.DependencyModel` and hence we need to pull in the latest here.

### DIFF
--- a/src/Dotnet.Script.DependencyModel.Nuget/Dotnet.Script.DependencyModel.NuGet.csproj
+++ b/src/Dotnet.Script.DependencyModel.Nuget/Dotnet.Script.DependencyModel.NuGet.csproj
@@ -14,6 +14,6 @@
     <Company>dotnet-script</Company>
   </PropertyGroup>
   <ItemGroup>
-    <PackageReference Include="Microsoft.CodeAnalysis.CSharp.Scripting" Version="2.6.0" />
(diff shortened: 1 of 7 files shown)


filipw:
thanks a lot for fixing this!

src/Dotnet.Script.DependencyModel.Nuget/Dotnet.Script.DependencyModel.NuGet.csproj
```diff
@@ -14,6 +14,6 @@
     <Company>dotnet-script</Company>
   </PropertyGroup>
   <ItemGroup>
-    <PackageReference Include="Microsoft.CodeAnalysis.CSharp.Scripting" Version="2.6.0" />
+    <PackageReference Include="Microsoft.CodeAnalysis.CSharp.Scripting" Version="2.6.1" />
```
filipw: actually, this only has the Mono hotfix so it won't affect us here

Issue #213, opened 2017-12-21
Let `build.sh` pick latest dotnet-script version programmatically


andmos:
I noticed the version of `dotnet-script` used by `build.sh` is set manually, which can cause some errors. (I see you, #208 🙈 ) We pick the latest version from the Github API in `install.sh`, might be worth stealing that curl 😄

filipw:
done

PR #217, opened 2018-01-02
picking latest version of dotnet-script when building. Fixing #213

andmos:


filipw:
@andmos thanks, any idea why this failed on macOS?

andmos:
Looking into it, worked locally on the macOS machines we have here...

andmos:
Found it, 
```
{
  "message": "API rate limit exceeded for 52.3.55.28. (But here's the good news: Authenticated requests get a higher rate limit. Check out the documentation for more details.)",
  "documentation_url": "https://developer.github.com/v3/#rate-limiting"
}
```
:( 
Makes sense, I would think Travis generates a lot of traffic without API tokens..

andmos:
@seesharper is adding some tokens now, Let's see if that fixes the issue!

seesharper:
@filip What if we simply create a dummy user on GitHub and create a personal access token for that user. That token can only be used to access repos belonging to the dummy user. This dummy user will not have any repos and the worst thing that could happen is that somebody steals the token and starts abusing the dummy user's account.

andmos:
Yeah that should work. The only thing we are after here is the Github API...

filipw:
sounds good to me

andmos:
The token did the trick. As long as we just have the dummy-user I think we should be good.

filipw:
thank you!

PR #219, opened 2018-01-09
Feature/218

seesharper:
This PR adds support for passing a custom filename during initialization.

```
dotnet script init custom.csx
```

This create a initialized folder with a script file named `custom.csx` 

```
dotnet script init
```

This will now create a file named `main.csx` instead of the old `HelloWorld.csx` file.

### DIFF
--- a/src/Dotnet.Script.Core/Scaffolder.cs
+++ b/src/Dotnet.Script.Core/Scaffolder.cs
@@ -1,5 +1,6 @@
 ﻿using System;
 using System.IO;
+using System.Linq;
 using System.Reflection;
 using Dotnet.Script.Core.Templates;
 
@@ -7,7 +8,7 @@ namespace Dotnet.Script.Core
 {
     public class Scaffolder
     {
-        public void InitializerFolder()
+        public void InitializerFolder(string fileName)
         {
             string currentDirectory = Directory.GetCurrentDirectory();
             string vsCodeDirectory = Path.Combine(currentDirectory, ".vscode");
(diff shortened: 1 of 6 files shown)


src/Dotnet.Script.Core/Scaffolder.cs
```diff
@@ -35,7 +36,13 @@ public void InitializerFolder()
                 WriteFile(pathToOmniSharpJson, omniSharpFileTemplate);
             }
 
-            CreateNewScriptFile("helloworld.csx");
+            if (Directory.GetFiles(currentDirectory, "*.csx", SearchOption.AllDirectories).Any()
```
filipw: is the depth-scan by design? I think we don't need to scan all subfolders, just the current one?

src/Dotnet.Script/Program.cs
```diff
@@ -88,10 +88,11 @@ private static int Wain(string[] args)
             app.Command("init", c =>
             {
                 c.Description = "Creates a sample script along with the launch.json file needed to launch and debug the script.";
+                var fileName = c.Argument("filename", "The name of the sample script file to be created during initialization.");
```
filipw: we can add info that it's optional

src/Dotnet.Script.Core/Scaffolder.cs
seesharper: Agreed. Top level should be enough 👍

src/Dotnet.Script/Program.cs
seesharper: Yup 👍

seesharper:
@filipw Are we good here ? 😃

filipw:
sorry I forgot :)

PR #223, opened 2018-02-08
Remove Global.json

Dispersia:
Global.json is no longer used with csproj's. Causes compatibility issues with users that only have SDK 2.1.4 installed, making it so you can't resolve back to 2.0.0 so you can't build.


filipw:
absolutely - thanks

Issue #44, opened 2016-12-15
Investigate running on full desktop CLR


filipw:
Mono/. NET 4.6

acormier:
I just now saw the discussion about DependencyContext.Default returning null here: https://github.com/filipw/dotnet-script/pull/224

filipw:
I think you are on some really old version, because that code doesn't exist for quite a while now.
That said, the current version of `Dotnet.Script.Core` would not work on desktop framework for at least two reasons:

1. there is code which tries to determine a current runtime identifier and if it's not `netcoreapXXX` it fails
2. the default references (such as `System.dll`, `System.Core.dll` and so on) are not implicitly added. That is because we pick them up from `project.assets.json`. This works fine on .NET Core, but on Desktop .NET the GAC references would be skipped and would have to be added manually

I am not sure if these are all the roadblocks, but these are definitely there from the top of my head.
There is no reason why these couldn't be fixed though, in fact this should be fairly uncomplicated to do so if this is something you feel like you'd need, then we could fix that.

acormier:
I had 0.21.0 checked out which is 10 days old.  I guess that is really old :-)

Our app hosts the .net framework and supporting C# scripting with dotnet-script is what I'm trying to do.  I can imagine that others would benefit from using Dotnet.Script.Core from Desktop apps

filipw:
> I guess that is really old :-)

😄

PR #224, opened 2018-02-09
Remove dependency on System.Runtime.Loader.

SlowLogicBoy:
Fixes runtime error when running on net framework
Unable to load System.Runtime.Loader

I don't know if this has any side effects :?

### DIFF
--- a/src/Dotnet.Script.Core/Dotnet.Script.Core.csproj
+++ b/src/Dotnet.Script.Core/Dotnet.Script.Core.csproj
@@ -36,7 +36,6 @@
     <PackageReference Include="Newtonsoft.Json" Version="10.0.3" />
     <PackageReference Include="System.Collections.Immutable" Version="1.4.0" />
     <PackageReference Include="System.Reflection.Metadata" Version="1.5.0" />
-    <PackageReference Include="System.Runtime.Loader" Version="4.3.0" />
     <PackageReference Include="System.ValueTuple" Version="4.4.0" />
   </ItemGroup>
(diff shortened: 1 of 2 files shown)


seesharper:
When you say "running on net framework" what does that mean exactly ? 
Do you put `#!  "net462"` or something similar at the beginning of the script?

SlowLogicBoy:
No, I add dependency on Dotnet.Script.Core in my project (.csproj) that targets net461, and when I try running script it throws exception.

seesharper:
I see. Just out of curiosity, what is the use case here?

SlowLogicBoy:
I want to run some Wpf stuff, and it runs only with net framework :)
Basically trying to solve #44 at least partially.

seesharper:
> Anyway that's not the only problem I have, trying to figure out other problems, like DependencyContext.Default == null for some reason

That is not going to be a simple thing to solve. We use something called a `DependencyContextReader` under the hood that parses the generated `project.assets.json` file and that does not work for full framework projects.

filipw:
We can merge it - it's one dependency less, and while .NET Core doesn't support everything related to AppDomains (i.e. create new) this stuff here works just fine with .NET Core 2.0.

That said, as @seesharper mentioned, to target desktop .NET, a lot of other changes would be needed unfortunately.

SlowLogicBoy:
Yea, after checking it out, it seems so.
For now I am able to run things I need on net core, I guess we'll see about the future :).

PR #229, opened 2018-03-01
Feature/release configuration

seesharper:
This PR adds support for specifying the debug/release configuration when executing a script.
The default is `debug` as before, but we can now specify the configuration to use like this

```
dotnet script -c release foo.csx
```

I also took the liberty to clean up some of the code that was related to `debugMode` which was passed around all over the place when all it is needed for is to determine the logging verbosity.

Note: Nothing has been done explicitly to support this in REPL mode. Running scripts in release mode is sort of an edge case and IMO not really needed in the REPL.

### DIFF
--- a/src/Dotnet.Script.Core/Interactive/InteractiveRunner.cs
+++ b/src/Dotnet.Script.Core/Interactive/InteractiveRunner.cs
@@ -38,9 +38,9 @@ namespace Dotnet.Script.Core
             _globals = new InteractiveScriptGlobals(Console.Out, CSharpObjectFormatter.Instance);
         }
 
-        public virtual async Task RunLoop(bool debugMode)
+        public virtual async Task RunLoop()
         {
-            while (true && !_shouldExit)
+            while (!_shouldExit)
             {
                 Console.Out.Write("> ");
                 var input = ReadInput();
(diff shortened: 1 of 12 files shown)


seesharper:
Obviously I broke something. Not at all related to the changes made, but still 😄 
Looking into it 👍

src/Dotnet.Script.Core/Interactive/InteractiveRunner.cs
```diff
@@ -38,9 +38,9 @@ public InteractiveRunner(ScriptCompiler scriptCompiler, ScriptLogger logger, Scr
             _globals = new InteractiveScriptGlobals(Console.Out, CSharpObjectFormatter.Instance);
         }
 
-        public virtual async Task RunLoop(bool debugMode)
+        public virtual async Task RunLoop()
         {
-            while (true && !_shouldExit)
+            while (!_shouldExit)
```
filipw: this was a left over from when it was only `while(true)` - thanks

src/Dotnet.Script.Tests/xunit.runner.json
```diff
@@ -1,3 +1,3 @@
 {
-  parallelizeTestCollections false
+  "parallelizeTestCollections":  false
```
filipw: 😃

filipw:
thanks for cleaning up that ugly `bool`

seesharper:
@filipw  Finally all green 👍

filipw:
🚀🚀

Issue #235, opened 2018-03-11
Assembly redirection doesn't check if assembly is already loaded.


SlowLogicBoy:
When dotnet script tries to redirect assembly, it doesn't check if assembly is already loaded.
After redirection I get same assembly loaded multiple times.

Also `Assembly.LoadFile()` doesn't load in CurrentContext, we should probably use `Assembly.LoadFrom(...)` ?

https://github.com/filipw/dotnet-script/blob/master/src/Dotnet.Script.Core/ScriptCompiler.cs#L194

PR #236, opened 2018-03-11
Fix for #235

SlowLogicBoy:
For me that If statement and return was enough. 
I also changed `Assembly.LoadFile()` to `Assembly.LoadFrom()`, because:
https://stackoverflow.com/a/1477899
I've ran into this before, but I can't recall Microsoft's article about Assembly loading, where it explains when to use `LoadFile` and `LoadFrom`

TODO:
- [x] Tests


filipw:
> I also changed Assembly.LoadFile() to Assembly.LoadFrom()

this is a very good change - thanks for spotting. Actually we also changed it OmniSharp recently

SlowLogicBoy:
Now I'm thinking if I even need that if, because LoadFrom() will return already loaded one anyway... Unless different version is loaded :?

seesharper:
@SlowLogicBoy Do you have a script that requires these changes? Would love to try it out and see where it fails in the current version

SlowLogicBoy:
I do but I don't want to share it, I'll have to figure out another way how to reproduce it.

seesharper:
@SlowLogicBoy Sounds good. Try to come up with the simplest thing that does not work 😸

SlowLogicBoy:
Difficult part is, I need to reference few dll's which reference nuget with different versions, and my script references latest one.

seesharper:
I see, maybe you can use an existing package on NuGet. Pick an older version of the package. Extract the contents of the package manually and reference the dll directly using the `#r` directive. In addition to this you reference the latest package using the `#r nuget` directive. It should give you the version mismatch you're looking for

seesharper:
Might seems like nit picking on our part, but I'm a little reluctant to applying a fix unless we know what we are fixing 😄

SlowLogicBoy:
It's understandable, I'm also strict when it comes to PR's for my projects :).

SlowLogicBoy:
It might be also a problem because I use DI, and when it's resolving types `LoadFile` happens.

SlowLogicBoy:
@seesharper, @filipw by any change do you know how to change "Submission#*" name?
Because I have a problem right here:
https://github.com/filipw/dotnet-script/pull/236/files#diff-b45a3e542b4d44b619a7ee12fb918c88R8

SlowLogicBoy:
For now I class TestClass with reflection, until we'll find proper solution.

src/Dotnet.Script.Tests/ScriptExecutionTests.cs
```diff
@@ -192,6 +200,25 @@ public void ShouldSupportInlineNugetReferencesWithTrailingSemicoloninEvaluatedCo
             Assert.Contains("AutoMapper.MapperConfiguration", result.output);
         }
 
+        [Fact]
+        public static void ShouldHandleIssue235()
```
filipw: I am not sure I understand what is this trying to test?
so it compiles a script into a DLL, and then runs a script that references that DLL - what conditions does that create? that certain DLL dependency is loaded twice?

filipw:
`Submission#` is hardcoded into Roslyn https://github.com/dotnet/roslyn/blob/version-2.7.0-beta3/src/Scripting/Core/ScriptBuilder.cs#L64

I don't think compiling a script into a DLL and then consuming it with intellisense and so on is something that Roslyn scripting ever wanted to support. 
You'd probably only consume script DLLs for non-automated scenarios like plugin discovery and for script re-reuns.

However if you have a use case, we should explore it.

src/Dotnet.Script.Tests/ScriptExecutionTests.cs
SlowLogicBoy: 1. In a script I call Mapper.CreateNew(), so that NuGet Assembly would be loaded,
2. I reference dependency dll, that also references that Nuget just with a different version
3. Execute referenced dlls method and dotnet-script tries to redirect outdated nuget assembly to a new one.
And the test is red before my changes.

I compile script to dll, because it was easiest solution for me to reproduce this behavior.

SlowLogicBoy: I know this isn't a clean solution :). I was hoping that if I do this you will point me in a better direction ;).

filipw: OK I see what you mean. Yeah there could be a simpler way to do it, but it's OK for me for now

src/Dotnet.Script.Core/ScriptCompiler.cs
```diff
@@ -190,8 +190,12 @@ private Assembly MapUnresolvedAssemblyToRuntimeLibrary(IDictionary<string, Runti
             {
                 if (runtimeAssembly.Name.Version > assemblyName.Version)
                 {
-                    Logger.Log($"Redirecting {assemblyName} to {runtimeAssembly.Name}");                    
-                    return Assembly.LoadFile(runtimeAssembly.Path);
+                    var loadedAssembly = AppDomain.CurrentDomain.GetAssemblies()
```
filipw: we will probably need to find a more efficient way to do it; it should be possible to attach to the `Resolving` event of `AssemblyLoadContext` and keep track of loaded assemblies rather than enumerate all the time

seesharper: If I remember correctly, the `MapUnresolvedAssemblyToRuntimeLibrary` does not get executed very often. Currently we have only one test that fails if we omit the handler all together. Not being a "hot path" I'll say that we can live with the enumeration for now 👍

SlowLogicBoy: @filipw do you want me to do something on this one?

src/Dotnet.Script.Tests/ScriptExecutionTests.cs
seesharper: @SlowLogicBoy , @filipw  Maybe a simpler approach would be to use the `dotnet cli` to create your test dll?

Something like 
```
using (var disposableFolder = new DisposableFolder())
{
    var projectFolder = Path.Combine(disposableFolder.Path, "TestLibrary");
    ProcessHelper.RunAndCaptureOutput("dotnet", new[] {"new classlib -n TestLibrary"}, disposableFolder.Path);
    ProcessHelper.RunAndCaptureOutput("dotnet", new[] { "add TestLibrary.csproj package AgileObjects.AgileMapper -v 0.23.0" }, projectFolder);
    ProcessHelper.RunAndCaptureOutput("dotnet", new[] {"build -c release -o ./"}, projectFolder);
}
```

SlowLogicBoy: I was thinking about making and pushing .csproj itself, but I couldn't find appropirate directory for it, maybe this disposable folder is a good idea, but wouldn't it be cheaper to just create csproj and push it, instead of creating a new one everytime?

seesharper: Performance in the test project is not all that important so you should be good creating it every time 👍

seesharper: @SlowLogicBoy Take a look at this. I think this pretty much tests the same as the original code.

```
public static void ShouldHandleIssue235()
{
    string code = 
    @"using AgileObjects.AgileMapper;
    public class TestClass
    {
        public TestClass()
        {
            IMapper mapper = Mapper.CreateNew();
        }
    }";

    string script =
    @"#! ""netcoreapp2.0""
    #r ""nuget: AgileObjects.AgileMapper, 0.23.1""
    #r ""TestLibrary.dll""
    
    using AgileObjects.AgileMapper;

    IMapper mapper = Mapper.CreateNew();
    var testClass = new TestClass();
    Console.WriteLine(""Hello World!"");";


    using (var disposableFolder = new DisposableFolder())
    {
        var projectFolder = Path.Combine(disposableFolder.Path, "TestLibrary");
        ProcessHelper.RunAndCaptureOutput("dotnet", new[] {"new classlib -n TestLibrary"}, disposableFolder.Path);
        ProcessHelper.RunAndCaptureOutput("dotnet", new[] { "add TestLibrary.csproj package AgileObjects.AgileMapper -v 0.23.0" }, projectFolder);
        File.WriteAllText(Path.Combine(projectFolder, "Class1.cs"), code);
        File.WriteAllText(Path.Combine(projectFolder, "script.csx"), script);
        ProcessHelper.RunAndCaptureOutput("dotnet", new[] {"build -c release -o ./"}, projectFolder);

        var dotnetScriptArguments = GetDotnetScriptArguments(Path.Combine(projectFolder, "script.csx"));
        var result = ProcessHelper.RunAndCaptureOutput("dotnet", dotnetScriptArguments);
        Assert.Contains("Hello World!", result.output);
    }
}
```

SlowLogicBoy: 👍 This works and way cleaner thanks!

SlowLogicBoy:
Crap sorry about the formatting changes on the last commit...

src/Dotnet.Script.Core/ScriptCompiler.cs
```diff
@@ -190,8 +190,12 @@ private Assembly MapUnresolvedAssemblyToRuntimeLibrary(IDictionary<string, Runti
...
+                        .FirstOrDefault(a => a.GetName().Name == assemblyName.Name);
+                    if(loadedAssembly != null)
```
filipw: one final thing, for the sake of understanding what is going on, let's log here a message that we redirect to an already loaded assembly

filipw:
thanks!

PR #233, opened 2018-03-06
Clear MSBuild related env variables before dotnet restore

seesharper:
This PR fixes #230. We need to get another release out the door so OmniSharp can reference the new `Dotnet.Script.DependencyModel` package. 👍

Issue #231, opened 2018-03-06
PowerShell installation script not working


fmajestic:
Copy-pasting the command gives these two errors:
```
PS C:\Users\Filip> (new-object Net.WebClient).DownloadString("https://raw.githubusercontent.com/filipw/dotnet-script/mas
ter/install/install.ps1") | iex


    Directory: C:\Users\Filip\AppData\Local\Temp


Mode                LastWriteTime         Length Name
----                -------------         ------ ----
d-----        6.3.2018.     16:10                dotnet-script
Invoke-WebRequest : The request was aborted: Could not create SSL/TLS secure channel.
At line:6 char:18
+ ... stRelease = Invoke-WebRequest "https://api.github.com/repos/filipw/do ...
+                 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : InvalidOperation: (System.Net.HttpWebRequest:HttpWebRequest) [Invoke-WebRequest], WebException
    + FullyQualifiedErrorId : WebCmdletWebResponseException,Microsoft.PowerShell.Commands.InvokeWebRequestCommand

Downloading latest verson ()
Exception calling "DownloadFile" with "2" argument(s): "The request was aborted: Could not create SSL/TLS secure channe
l."
At line:16 char:1
+ $client.DownloadFile($url,$zipFile)
+ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : NotSpecified: (:) [], MethodInvocationException
    + FullyQualifiedErrorId : WebException

Microsoft.PowerShell.Archive\Expand-Archive : The path 'C:\Users\Filip\AppData\Local\Temp\dotnet-script\dotnet-script.zip' either does not exist or is not a valid file system path.
At line:19 char:1
+ Microsoft.PowerShell.Archive\Expand-Archive $zipFile -DestinationPath ...
+ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : InvalidArgument: (C:\Users\Filip\...tnet-script.zip:String) [Expand-Archive], InvalidOperationException
    + FullyQualifiedErrorId : ArchiveCmdletPathNotFound,Expand-Archive
```

PowerShell version:
```
[...]

Issue #230, opened 2018-03-05
Intellisense not working for inline dependencies


vincentlabatut:
Hi, 
I've been reloading visual studio code many times, I'm trying to do as in the doc (https://github.com/filipw/dotnet-script#nuget-packages) but intellisense would not work with the following piece of code :
```
#! "netcoreapp2.0"

#r "nuget: Newtonsoft.Json,11.0.1"
#load "simpleHttpRequest.csx"

using System.Linq;
using Newtonsoft.Json.Linq;
```
The script works at runtime but in Visual Studio code I have Newtonsoft underlined in red, and no intellisense. (the same as for all JObject uses etc.)

VS Code 1.20.1
ms-vscode.csharp : 1.14.0
dotnet.script : 0.18.0

seesharper:
Is this on Windows?

SlowLogicBoy:
There are few caveats:
1. Run dotnet script init 
    it generates omnisharp.json which adds option 
   ```json
   "script": {
    "enableScriptNuGetReferences": true
   },
   ```
    only with this option omnisharp can recognize nugets referenced by `#r "nuget: ..."`

2. Everytime you add new `#r` you have to restart omnisharp because it doesn't automatically catch them.

vincentlabatut:
Hi, 
@seesharper : yes it is on Windows

@lowLogicBoy : Thanks. Actually, the files initial structure had been created with "dotnet script init", the omnisharp.json file already contains :
```
{
  "script": {
    "enableScriptNuGetReferences": true
  }
}
```
Restarting omnishart or restarting VS Code itself does not change the situation...

SlowLogicBoy:
Do you have `dotnet sdk 2.1.4` installed?
Can you provide omnisharp log?

seesharper:
@vincentlabatut In VS Code (Task tab) select `OmniSharp Log`.

vincentlabatut:
Yes this the version I have. 2.1.4

Attached omnisharp log

[omnisharp.txt](https://github.com/filipw/dotnet-script/files/1780887/omnisharp.txt)

seesharper:
This is where it fails
```
[fail]: OmniSharp.Script.ScriptProjectSystem
        Failed to resolve compilation dependencies
```

Actually, I have the exact same problem, but that is on a Mac. Might be related though.

Could you also enable detailed logging from OmniSharp?

User Settings (Not omnisharp.json)

```
{
[...]

Issue #221, opened 2018-01-16
Is possible to run script in RELEASE mode?


wk-j:
I'm trying to execute following code with dotnet script but it seem like BenchmarkDotNet not allow running benchmarks in DEBUG mode.

```csharp
#! "netcoreapp2.0"
#r "nuget:NetStandard.Library,2.0"
#r "nuget:BenchmarkDotNet,0.10.12"
#r "nuget:System.Memory,4.4.0-preview1-25305-02"

using BenchmarkDotNet.Attributes;
using BenchmarkDotNet.Running;

[MemoryDiagnoser]
public class Test {

    int[] values = new[] { 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12 };

    [Benchmark]
    public void Take() {
        var a = values.Skip(5).Count();
    }

    [Benchmark]
    public void Slice() {
        var span = new Span<int>(values, 0);
        var a = span.Slice(5, span.Length - 5).Length;
    }
}

BenchmarkRunner.Run<Test>();
```

```
❯ dotnet script "./scripts/Slice.csx"
// ***** BenchmarkRunner: Start   *****
// Found benchmarks:
//   Test.Take: DefaultJob
//   Test.Slice: DefaultJob

// Validating benchmarks:
Assembly ℛ*beb7aab2-3d06-4c0c-83f8-7cd9cf431b19#2-0 which defines benchmarks is non-optimized
Benchmark was built without optimization enabled (most probably a DEBUG configuration). Please, build it in RELEASE.
// * Artifacts cleanup *
```

seesharper:
H @wk-j 
There is currently no way to execute scripts in release mode. 
@filipw We should maybe add an option for this. It should be interesting to see if that affects the startup time. 

Something like 
```
-c, --configuration <CONFIGURATION>   Configuration to use for running the script [Release/Debug] Default is "Debug".
```

This way we can also add this to the `launch.json` file so that <kbd>CTRL</kbd>+<kbd>F5</kbd> (Start without debugging) adds the `-c release` option. That way we can execute in release mode from within VS Code

filipw:
It was there, but we removed it. However, it didn't behave as expected anyway, so we might as well say it has never been there 😀

The reason for not having it, was really to align with CSI, which also doesn't have it, and emit symbols all the time. We can add it as an opt-in though.

filipw:
[...]

PR #237, opened 2018-03-12
Build/continuous deployment

seesharper:
This PR adds support for continuous deployment so that we can release with less effort and without the risk of human errors during packaging and deployment. 

Deployment is triggered by creating a new tag on the master branch and pushing the tags to the origin.

The following example shows how to create a new release.

```
git checkout master 
git pull
git tag <new tag>
git push --tags
```

> The two first command are just to ensure that we have the latest from origin before creating the tag. 

The build script detects that we are on the master branch by checking the `BuildEnvironment.IsSecure` property. More on this later.

The build script then checks to see if this is a tag commit and if that is the case, the deployment process begins. 

For `dotnet-script` this means the following steps.

- Produce a zip file of the publish folder (release). ex `dotnet-script.0.19.0.zip`

- Create a release note to be published as part of the GitHub release.

- Produce a Chocolatey package. ex `dotnet.script.0.18.0.nupkg`

- Produce NuGet packages for `Dotnet.Script`, `Dotnet.Script.Core`, `Dotnet.Script.DependencyModel` and `Dotnet.Script.DependencyModel.NuGet`

  ​

When the artifacts has been created the build script will deploy these artifacts to their respective destinations. The zip file along with the release note goes to GitHub releases, the NuGet packages to NuGet and finally the Chocolatey package goes to Chocolatey. 

Artifacts are also uploaded to the AppVeyor  cloud service in case we need to point somebody to a specific build of some package not yet released. 

As mentioned the Release note is automatically generated and will be posted to GitHub Releases as the default release note. This is merely just a suggestion as to the content of the release note, but it should if not usable as it is, provide enough information about the changes that are included in the release. 

For instance this is how it would have looked like for the 0.19.0 release. 

# Change Log

### [Full Changelog](https://github.com/filipw/dotnet-script/compare/0.18.0...664f3204ff211ed623860b00a231f3e673f204ea)

**Merged Pull Requests**

- version bump to 0.19, build using 0.18 (22.01.2018) [\#222](https://github.com/filipw/dotnet-script/pull/222) ([filipw](https://github.com/filipw))
- Remove Global.json (14.02.2018) [\#223](https://github.com/filipw/dotnet-script/pull/223) ([Dispersia](https://github.com/Dispersia))
- Remove dependency on System.Runtime.Loader. (16.02.2018) [\#224](https://github.com/filipw/dotnet-script/pull/224) ([SlowLogicBoy](https://github.com/SlowLogicBoy))
- Feature/release configuration (02.03.2018) [\#229](https://github.com/filipw/dotnet-script/pull/229) ([seesharper](https://github.com/seesharper))
- Enforce TLS 1.2 in powershell (06.03.2018) [\#232](https://github.com/filipw/dotnet-script/pull/232) ([filipw](https://github.com/filipw))
- Clear MSBuild related env variables before dotnet restore (06.03.2018) [\#233](https://github.com/filipw/dotnet-script/pull/233) ([seesharper](https://github.com/seesharper))

**Closed Issues**

- support custom filename in init (10.01.2018) [\#218](https://github.com/filipw/dotnet-script/issues/218) ([filipw](https://github.com/filipw))
- Very large delays between command entry and execution (18.01.2018) [\#220](https://github.com/filipw/dotnet-script/issues/220) ([nesteruk](https://github.com/nesteruk))
- Is possible to run script in RELEASE mode? (02.03.2018) [\#221](https://github.com/filipw/dotnet-script/issues/221) ([wk-j](https://github.com/wk-j))
- Intellisense not working for inline dependencies (06.03.2018) [\#230](https://github.com/filipw/dotnet-script/issues/230) ([vincentlabatut](https://github.com/vincentlabatut))
- PowerShell installation script not working (06.03.2018) [\#231](https://github.com/filipw/dotnet-script/issues/231) ([fmajestic](https://github.com/fmajestic))

--> End of release note sample

The release note is generated using [github-changelog](https://github.com/seesharper/github-changelog) which is a script package that I have been working on for some time now. 

Another script package, [dotnet-build](https://github.com/seesharper/dotnet-build) simply contains a collection of build related script such as `NuGet.csx`, `Command.csx`, `FileUtils.csx` and so on. 

In order for all this to work properly, we need to provide the necessary encrypted environment variables to AppVeyor.

This is the `appveyor.yml` files as is stands in this PR.

```yaml
environment:
    IS_SECURE_BUILDENVIRONMENT:
        secure: The encoded value of 'true' (not including the apostrophes)
    GITHUB_REPO_TOKEN:
        secure: The encoded value of your GitHub Access Token.
    NUGET_APIKEY:
        secure: The encoded value of your NuGet ApiKey.
    CHOCO_APIKEY:
        secure: The encoded value of the Choco API Key (dotnet-script account).
```

These values are ONLY decoded when a build is started from the `master` branch and hence the `IS_SECURE_BUILDENVIRONMENT` will only evaluate to `true` for the master branch. 

You might wonder why we don't simply check if we are on the master branch? The reason is that AppVeyor never builds anything directly of the master branch. It build from a particular commit provided in the web hook. The reason for not checking the AppVeyor specific environment variables is that I wanted the build script to be completely unaware of the underlying build server. 

The other three variables are just access keys to the deployment targets.

For local builds running on your machine, these values as just present as environment variables in plain text. 

So, I need you (@filipw ) to encrypt these values using the `Encrypt Data` feature of AppVeyor.  

Even though I do have access to AppVeyor, I am not entirely sure how it would be if I encrypted the variables. Figured best if we did it from one account (yours:))

> I'll mail you the password for the `dotnet-script` account on Chocolatey so you can get the API key needed for pushing to Chocolatey.- 

 

With this in place we'd never need to do a  "manual" again apart from maybe adding some TLC to the generated release notes. That's it.


build/Choco.csx
```diff
@@ -1,6 +1,4 @@
-#load "FileUtils.csx"
-#load "DotNet.csx"
-#load "Command.csx"
+#load "nuget:Dotnet.Build, 0.2.5"
```
filipw: 🏆

filipw:
Looks great! Looks like we will avoid publishing a wrong dll like I did last time 😂

> As mentioned the Release note is automatically generated and will be posted to GitHub Releases as the default release note.

doesn't this also create a tag?

Also, given that Nuget doesn't allow packages to be deleted, I think I'd like to run it once without auto-publish and just create artifacts, and only from then onwards enable auto-publish. Would that make sense?

seesharper:
> doesn't this also create a tag?

Not exactly. A tag and a release are two different things. A GitHub release is its own thing that is associated with a given tag. In this case it is the latest tag.

That being said, for 0.19.0 there is already an existing draft that probably should put aside while we run this for the first time. We'll coordinate that when are ready to release. No worries there 😄 

> Also, given that Nuget doesn't allow packages to be deleted, I think I'd like to run it once without auto-publish and just create artifacts, and only from then onwards enable auto-publish. Would that make sense?

In the `dotnet-script` solution we produce multiple NuGet packages, but we don't necessarily bump the version for all the packages. That is why you see the `NuGet.TryPush` instead of `NuGet.Push` which would fail if it tries to publish an existing package.  

We should be good with the existing stuff in the build script. The only missing piece is the various API keys that needs to be encrypted and put into the `appveyor.yml` file. 

Drop me a mail at bernhard.richter@gmail.com and I'll provide the account information for Chocolatey 👍

filipw:
> Not exactly. A tag and a release are two different things. A GitHub release is its own thing that is associated with a given tag. In this case it is the latest tag.

sorry for not being clear. I meant that if I go to releases on GH, I can create a new release and give it a tag. That tag is then publicly created when I publish the release.
So what happens, if I do something like this now? It seems that the script would run on appveyor and create a new release draft?

Or in other words, I could trigger the release process via GH release UI, without dropping to command line at all, which is even more enticing scenario I would say.

seesharper:
Ah, okay. I understand. That should be quite easy to accomplish. Just need to check if the AppVeyor build is triggered when we create a tag from the GitHub UI. If that works we should be okay. I also need to check the logic for actually creating the release when there is already a release associated with the tag. I'll get back at you 👍

seesharper:
Confirmed. The AppVeyor build is triggered when creating the tag from the GitHub UI. Same as `git push --tags` Just need to check if we can update the release notes and add the assets 👍

filipw:
great. let's do this release 0.19.0 manually since I already drafted it last week and then from next one we automate everything 👍

seesharper:
Sounds good to me. Creating the release from the GitHub UI totally possible. Making the necessary changes now 👍

seesharper:
With that we can release either from the command line or from the GitHub Release UI 👍  Fully automated

filipw:
I have released 0.19.0 manually, can you push the choco release? then we can merge this and move to a more structured world of managing releases!

seesharper:
The choco package has been pushed and is awaiting moderation. Should be available soon. I'll need to make some minor changes to this PR to make the GitHub UI scenario work. Let you know once it's ready 👍

seesharper:
I've made the necessary changes in `Dotnet.Build` to handle creating a new tag/release from the GitHub UI. 
The workflow will be.
1. Create a new release with a new tag in the GitHub Releases UI.
2. Leave the name and body empty as it will be overwritten by the build script that pushes the generated changelog.
3. After the AppVeyor build completes and the release notes have been updated, we are free to edit the release if needed. 

> Note: `git push --tags` will also work if we want to do it from the command line

**Important**
@filipw I need you to fill inn the missing keys in the `AppVeyor.yml` file before we merge to master.
You have all the information needed except for the Choco api key. Ping me on twitter and we'll exchange that info somehow. 👍 

Since there are parts of the build script that only executes when building on master, we might encounter issues as the code has never been executed before. If that happens, I'll fix it immediately. That would require an "in-flight push to master". Nothing to worry about. I have implemented this on several of my other projects here on GitHub and its working perfectly 👍

filipw:
ok let's go

Issue #226, opened 2018-02-25
DLL/Type Loading error


SlowLogicBoy:
Strange behavior can't figure out what's happening.
What I know is that when nuget that I reference, is dependent on other version of already loaded assembly it no longer can load types.

Only happens on my Linux machine, on Windows everything is fine.

Code:
```csharp
#! "netcoreapp2.0"
#r "nuget: Castle.Core,*"

Console.WriteLine("Hello");
Process.Start("dotnet", "--version").WaitForExit();
Console.WriteLine("End");
```
```
user@user-pc ~/code/Test $ dotnet script --version

0.18.0
user@user-pc ~/code/Test $ dotnet --version
2.1.4
user@user-pc ~/code/Test $ dotnet script main.csx
/home/user/code/Test/main.csx(5,9): error CS7069: Reference to type 'Component' claims it is defined in 'System.ComponentModel.Primitives', but it could not be found
user@user-pc ~/code/Test $ dotnet script
> #load "main.csx"
Hello
2.1.4
End
>
```

Other nugets that also yield this error:
```
#r "nuget: Foundatio.Repositories, *"
#r "nuget: YoutubeExplode,*"
```
Loaded Assemblies are the same both ways.

seesharper:
What you are saying here is that if you load the script via the REPL it works?

SlowLogicBoy:
That's right.

seesharper:
@SlowLogicBoy Is this fixed by #235  or possibly related?

SlowLogicBoy:
Nope, #235 did not fix this one.
Is it related? Can't tell, I think this one is something to do Linux version of dotnet, because this doesn't happen on windows machines.

seesharper:
@SlowLogicBoy 
I just tried this on my Ubuntu 16.04 machine and I am unable to reproduce this error.

```
#! "netcoreapp2.0"
#r "nuget: Castle.Core,*"

Console.WriteLine("Hello");
Process.Start("dotnet", "--version").WaitForExit();
Console.WriteLine("End");
```

yields the following output 

```
Hello
2.1.300-preview1-008174
End
```

Environment:

```
dotnet-script version : 0.20.0

Product Information:
 Version:            2.1.300-preview1-008174
 Commit SHA-1 hash:  b8df89a54f

Runtime Environment:
 OS Name:     ubuntu
 OS Version:  16.04
 OS Platform: Linux
 RID:         ubuntu.16.04-x64
[...]

PR #238, opened 2018-03-20
Issue226 test.

SlowLogicBoy:


seesharper:
Is this the test proving failure from #226 ?

SlowLogicBoy:
Yea, but the test is inaccurate, fixing that now.

SlowLogicBoy:
Looks like travis-ci Linux build succeeded, I guess then my environment setup is incorrect.

src/Dotnet.Script.Tests/TestFixtures/Issue226/Issue226.csx
```diff
@@ -0,0 +1,12 @@
+#! "netcoreapp2.0"
+#r "nuget: Castle.Core,*"
+
+Process.Start(new ProcessStartInfo {
```
seesharper: Is it referencing `Castle.Core` that causes the error is it starting the process? And what is the relation here? 😄

SlowLogicBoy: The error happens only when I reference one of "those" nugets, that probably reference ether new or older `System.ComponentModel.Primitives` package, that would be my guess.

SlowLogicBoy: I know there were some kind of a bug when you upgrade I think from dotnet 2.0.* to 2.1.* that might be my problem. I already reinstalled it on my machine quite a few times, but I think there was another step that I needed todo, however can't find the article for this -.-.

seesharper: Does this mean that it still fails for you on Linux?

SlowLogicBoy: Yes.

seesharper: Are you running .Net Core 2.1 Preview?

SlowLogicBoy:
```
me@my-pc ~ $ apt list --installed | grep dotnet*

WARNING: apt does not have a stable CLI interface. Use with caution in scripts.

dotnet-host/xenial,now 2.1.0-preview1-26216-03-1 amd64 [installed,automatic]
dotnet-hostfxr-2.0.5/xenial,now 2.0.5-1 amd64 [installed,automatic]
dotnet-runtime-2.0.5/xenial,now 2.0.5-1 amd64 [installed,automatic]
dotnet-runtime-deps-2.1.0-preview1-26216-03/xenial,now 2.1.0-preview1-26216-03-1 amd64 [installed,automatic]
dotnet-sdk-2.1.4/xenial,now 2.1.4-1 amd64 [installed]
```
I'm pretty sure I do :)

seesharper: I see. The Travis build uses .net Core 2.0 so that is probably why it does not fail there. 
We haven't really started to look at .Net Core 2.1 yet, but I guess it would be possible to add that to the build matrix as well. It does not make all that much sense to merge this without seeing it fail. Only then would we know where to fix it 👍

seesharper:
@filipw Should we set up a build based on .Net Core 2.1 preview or do we wait until RTM? Guess it does not hurt to be a little prepared for what's coming 😄

src/Dotnet.Script.Tests/TestFixtures/Issue226/Issue226.csx
SlowLogicBoy: Hmm, haven't thought that 2.1 might be a problem, I might try downgrading to 2.0 :). It's not a big deal, I made this PR, so it would be easier to see where the problem is, but since there is no problem we can close this PR.

SlowLogicBoy:
I can try updating travis-ci with net sdk 2.1 preview on this pr to see if it fails :?

seesharper:
> I can try updating travis-ci with net sdk 2.1 preview on this pr to see if it fails :?

I think we just wait and see if thus is a problem when 2.1 ships. 

If you just update the branch, we'll merge this as a regression test ensuring that we don't hit this in the future

SlowLogicBoy:
I've reinstalled dotnet and removed fallback packages, tried with dotnet sdk 2.0.3 and 2.1.4 still no luck.
And I can clearly see while debugging that there is `System.ComponentModel.Primitives 4.2.0.0` loaded and that assembly contains `Component` type, so I guess something is still wrong with Assembly loading, at least that would be my guess.

SlowLogicBoy:
The important thing is that I can run `Process.Start` in `.csproj` projects and in Tests themselves (well none of the tests would work), and when in `dotnet script` interactive, but not with `dotnet script Issue226.csx`

PR #245, opened 2018-04-10
set .net cli environment variables in CI

filipw:
Improve CI build performance.
Do not build up package cache + do not send telemetry from CI.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,9 @@
 language: csharp
 dotnet: 2.0.0
+env:
+  global:
+    - DOTNET_SKIP_FIRST_TIME_EXPERIENCE=true
+    - DOTNET_CLI_TELEMETRY_OPTOUT=true
 mono:
  - latest
 matrix:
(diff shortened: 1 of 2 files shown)


atifaziz:
> Improve CI build performance.

Is this measurable? I'm curious to know if it made a remarkable impact on build times (as in going from X to Y minutes where Y < X).

atifaziz:
@filipw Thanks and I was precisely looking for that text quoted by @DarqueWarrior on his blog post:

> `A command is running to initially populate your local package cache, to improve restore speed and enable offline access. This command will take up to a minute to complete and will only happen once.`

I didn't find it on previous builds in the build history ([AppVeyor](https://ci.appveyor.com/project/filipw/dotnet-script/history)/[Travis](https://travis-ci.org/filipw/dotnet-script/builds)), taking a [recent build](https://ci.appveyor.com/project/filipw/dotnet-script/build/1.0.445) for b3901973e5f7db9ea78e2e63b4714d4ed8155684 as an example. Then I would have expected that it disappeared in the build for this PR. It could be that the images used AppVeyor and Travis have the cache primed so those build are not considered first-time runs.

PR #248, opened 2018-04-12
Scaffolding/default target framework

seesharper:
This PR adds support for setting the `defaultTargetFramework` during `dotnet init`

It now produces the following `omnisharp.json`

```
{                                            
  "script": {                                
    "enableScriptNuGetReferences": true,     
    "defaultTargetFramework": "netcoreapp2.0"
  }                                          
}                                            
```
Also bumped the version to 0.21.0 and fixed the build script so that it pushes the package to Chocolatey.

### DIFF
--- a/src/Dotnet.Script.Core/Templates/omnisharp.json.template
+++ b/src/Dotnet.Script.Core/Templates/omnisharp.json.template
@@ -1,5 +1,6 @@
 ﻿{
   "script": {
-    "enableScriptNuGetReferences": true
+    "enableScriptNuGetReferences": true,
+    "defaultTargetFramework": "netcoreapp2.0"
   }
 }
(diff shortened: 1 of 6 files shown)


src/Dotnet.Script.Core/Templates/omnisharp.json.template
```diff
@@ -1,5 +1,6 @@
 ﻿{
   "script": {
-    "enableScriptNuGetReferences": true
+    "enableScriptNuGetReferences": true,
+    "defaultTargetFramework": "netcoreapp2.0"
```
filipw: should we remove the directive from the scaffolded `main.csx`?
also, maybe this value could be passed from the scaffolder so that we don't need to manually maintain the template

seesharper: I think maybe we should wait with removing the directive until the next version of the C# extension for Visual Studio Code. If we remove the directive now we will break intellisense

seesharper: When the next release of OmniSharp is shipped via the C# extension we can change this. We can even do the *nix shebang when running on *nix. But until that happens, I think it would be best to leave it as it is

PR #250, opened 2018-04-13
Feature/remote

seesharper:
This PR adds support for executing script made available on an http/https endpoint 

Usage 
```
dotnet script https://tinyurl.com/y8cda9zt
```

### DIFF
--- a/src/Dotnet.Script.Core/ScriptDownloader.cs
+++ b/src/Dotnet.Script.Core/ScriptDownloader.cs
@@ -1,6 +1,7 @@
 ﻿using System;
 using System.IO;
 using System.Net.Http;
+using System.Net.Mime;
 using System.Threading.Tasks;
 
 namespace Dotnet.Script.Core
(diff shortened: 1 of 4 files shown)


src/Dotnet.Script.Core/ScriptDownloader.cs
```diff
@@ -0,0 +1,28 @@
...
+            using (HttpClient client = new HttpClient())
+            {
+                using (HttpResponseMessage response = await client.GetAsync(uri))
+                {
+                    response.EnsureSuccessStatusCode();
+                    using (HttpContent content = response.Content)
+                    {
+                        string scriptFile = $"{Path.GetTempFileName()}.csx";                        
+                        var code = await content.ReadAsStringAsync();
+                        File.WriteAllText(scriptFile, code);
```
filipw: do we need to write the file? we could - I think - just buffer it in memory and call `RunCode()` (the eval path)

filipw:
great!

src/Dotnet.Script.Core/ScriptDownloader.cs
filipw: actually, when we have a file, we could debug. however you'd need to know where the file is, so not sure how useful that is. Maybe in combination with the dotnet-script extension that could work.

the advantage of not having the file is no need to clean up anything. For example at the moment if anything throws, looks like the file would not be cleaned, so we should at least add that if we don't stay in memory

filipw:
LGTM

PR #253, opened 2018-04-14
use  #!/bin/bash instead of  #!/bin/sh

filipw:
We use bash syntax but specify #!/bin/sh. if that's symlinked elsewhere, i.e. to dash, it will show weird error messages.


seesharper:
Makes sense. Merging 👍

Issue #254, opened 2018-04-15
Missing runtime assemblies on Debian Stretch


filipw:
The famous issue #204 fails on Debian Stretch

```
root@70144b5a5421:/dotnet-script/src/Dotnet.Script.Tests# dotnet test
Build started, please wait...
Build completed.

Test run for /dotnet-script/src/Dotnet.Script.Tests/bin/Debug/netcoreapp2.0/Dotnet.Script.Tests.dll(.NETCoreApp,Version=v2.0)
Microsoft (R) Test Execution Command Line Tool Version 15.6.0
Copyright (c) Microsoft Corporation.  All rights reserved.

Starting test execution, please wait...
[xUnit.net 00:00:00.7570662]   Discovering: Dotnet.Script.Tests
[xUnit.net 00:00:00.8246965]   Discovered:  Dotnet.Script.Tests
[xUnit.net 00:00:00.8317680]   Starting:    Dotnet.Script.Tests
[xUnit.net 00:01:44.5950436]     Dotnet.Script.Tests.ScriptExecutionTests.ShouldHandleIssue204 [FAIL]
[xUnit.net 00:01:44.5990250]       Assert.Contains() Failure
[xUnit.net 00:01:44.5995086]       Not found: System.Net.WebProxy
[xUnit.net 00:01:44.5997346]       In value:  /dotnet-script/src/Dotnet.Script.Tests/bin/Debug/netcoreapp2.0/../../../TestFixtures/Issue204/Issue204.csx(5,26): error CS0246: The type or namespace name 'WebProxy' could not be found (are you missing a using directive or an assembly reference?)
[xUnit.net 00:01:44.6009356]       Stack Trace:
[xUnit.net 00:01:44.6021430]         /dotnet-script/src/Dotnet.Script.Tests/ScriptExecutionTests.cs(132,0): at Dotnet.Script.Tests.ScriptExecutionTests.ShouldHandleIssue204()
Failed   Dotnet.Script.Tests.ScriptExecutionTests.ShouldHandleIssue204
Error Message:
 Assert.Contains() Failure
Not found: System.Net.WebProxy
In value:  /dotnet-script/src/Dotnet.Script.Tests/bin/Debug/netcoreapp2.0/../../../TestFixtures/Issue204/Issue204.csx(5,26): error CS0246: The type or namespace name 'WebProxy' could not be found (are you missing a using directive or an assembly reference?)
Stack Trace:
[...]

PR #256, opened 2018-04-16
converted to multistage build in Dockerfile

andmos:
As mentioned in #254, this adds [multistage build](https://docs.docker.com/develop/develop-images/multistage-build/) to the `Dockerfile`.
In a nutshell, we first create a Docker image for building the project. This image includes Mono and NuGet for running all the tests. If all tests pass, we run `dotnet publish` in `release` mode and copy the artifacts to a simpler container, only containing the `dotnet-SDK`.
This makes the `Dockerfile` capable of doing the entire build (including NuGet restore) without adding unnecessary fluff in the final container image. 

Still:
```
cd build
docker build -t dotnet-script -f Dockerfile ..
```

### DIFF
--- a/build/Dockerfile
+++ b/build/Dockerfile
@@ -1,11 +1,24 @@
-FROM microsoft/dotnet:2.0.0-sdk 
+FROM microsoft/dotnet:2.0.0-sdk as builder
 COPY . /dotnet-script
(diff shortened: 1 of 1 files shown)


filipw:
thanks 👍

PR #261, opened 2018-04-21
feature/netcoreapp2.1

### LINKED ISSUE #258: Dotnet Global Tool Install - FileNotFoundException
Ran into an error when installing via `dotnet tool install` (letting dotnet grab the package from Nuget, nothing downloaded locally)

```
C:\test> dotnet tool install -g dotnet-script                                                                                                                    
Format version is missing. This tool may not be supported in this SDK version. Please contact the author of the tool.                                                           
You can invoke the tool using the following command: dotnet-script                                                                                                              
Tool 'dotnet-script' (version '0.21.0') was successfully installed.                                                                                                             
C:\test> dotnet script init                                                                                                                                      
                                                                                                                                                                                
Unhandled Exception: System.IO.FileNotFoundException: Could not load file or assembly 'C:\Users\xxx\.dotnet\tools\dotnet-script.dll'. The system cannot find the file specified.
```
Not sure why it is looking for a DLL and not the EXE that is in that directory.
I've tried this from powershell + cmd + restarted my sessions on both etc. with the same result.

System info:

```
C:\test> dotnet --info
.NET Core SDK (reflecting any global.json):
 Version:   2.1.300-preview2-008530
 Commit:    822ae6d43a

Runtime Environment:
 OS Name:     Windows
 OS Version:  10.0.17133
 OS Platform: Windows
 RID:         win10-x64
 Base Path:   C:\Program Files\dotnet\sdk\2.1.300-preview2-008530\

Host (useful for support):
  Version: 2.1.0-preview2-26406-04
  Commit:  6833f3026b

.NET Core SDKs installed:
  1.1.7 [C:\Program Files\dotnet\sdk]
  1.1.8 [C:\Program Files\dotnet\sdk]
  2.1.4 [C:\Program Files\dotnet\sdk]
  2.1.102 [C:\Program Files\dotnet\sdk]
  2.1.200-preview-007517 [C:\Program Files\dotnet\sdk]
  2.1.300-preview1-008174 [C:\Program Files\dotnet\sdk]
  2.1.300-preview2-008530 [C:\Program Files\dotnet\sdk]

.NET Core runtimes installed:
  Microsoft.AspNetCore.All 2.1.0-preview1-final [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.All]
  Microsoft.AspNetCore.All 2.1.0-preview2-final [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.All]
  Microsoft.AspNetCore.App 2.1.0-preview1-final [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.App]
  Microsoft.AspNetCore.App 2.1.0-preview2-final [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.App]
  Microsoft.NETCore.App 1.0.10 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
  Microsoft.NETCore.App 1.1.6 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
  Microsoft.NETCore.App 1.1.7 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
  Microsoft.NETCore.App 2.0.5 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
  Microsoft.NETCore.App 2.0.6 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
  Microsoft.NETCore.App 2.1.0-preview1-26216-03 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
  Microsoft.NETCore.App 2.1.0-preview2-26406-04 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
```

If you need any more info let me know.

seesharper:
This PR adds support for cross compiling `dotnet-script` using `netcoreapp2.0` and `netcoreapp2.1`
The motivation for this is to fix #258 that requires global tools to be compiled for at least `netcoreapp2.1`.

Please refer to this issue for form information https://github.com/dotnet/cli/issues/9073

In the process of adding support for netcoreapp2.1, it was important that existing scripts that contained the `#! "netcoreapp2.0"` pragma should continue to work.

This resulted in a deep dive into how we add assemblies (metadatareferences) for a given script.

We used to add a couple of "inherited" assembly references from the host (dotnet-script) and then load the rest from the dependency context inferred from the script. 

The initial hypothesis was that we needed to get rid of all inherited assembly references since there would be a possible mismatch between the host running as `netcoreapp2.1` and the script dependencies being resolved in the context of a `netcoreapp2.0` application. 

As it turns out when compiling and executing a script , the script does not at all run in 100% isolation from the host. For instance it shares the same AssemblyLoadContext so that means we can't really ignore the assemblies already loaded into the host AppDomain/AssemblyLoadContext.

So the approach taken is that we now inherit "on demand" meaning that when we load a assembly reference from the script dependency context, we will now check if the assembly is already loaded by the host. If its already loaded we skip the reference from the script dependency context, otherwise we load it. This should also fix the scenario when we run as `netcoreapp2.0` where we only have the ``` `netcoreapp2.1` sdk installed.

This change actually made all tests pass without suppressing `CS1705`. 

In this PR we still suppress certain diagnostics, but they are now logged when running in debug mode `-d`

When it comes to the packages we produce nothing much has changed.

The tool package is compiled for `netcoreapp2.1` while the rest of the packages are `netcoreapp2.0` as before

### DIFF
--- a/build/Build.csx
+++ b/build/Build.csx
@@ -100,7 +100,7 @@ private void PatchContent(string solutionFolder)
 {
     var pathToDotnetScriptProject = Path.Combine(solutionFolder,"Dotnet.Script","Dotnet.Script.csproj");
     var projectFile = XDocument.Load(pathToDotnetScriptProject);
-    var contentElements = projectFile.Descendants("Content");
+    var contentElements = projectFile.Descendants("Content").ToArray();
     foreach (var contentElement in contentElements)
     {
         contentElement.Remove();
--- a/build/Build.csx
+++ b/build/Build.csx
@@ -1,5 +1,5 @@
 #! "netcoreapp2.0"
-#load "nuget:Dotnet.Build, 0.2.9"
+#load "nuget:Dotnet.Build, 0.3.0"
 #load "nuget:github-changelog, 0.1.4"
 #load "Choco.csx"
 #load "BuildContext.csx"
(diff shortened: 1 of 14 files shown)


build/Build.csx
```diff
@@ -1,28 +1,48 @@
 #! "netcoreapp2.0"
-#load "nuget:Dotnet.Build, 0.2.9"
+#load "nuget:Dotnet.Build, 0.3.1"
```
filipw: what changed here?

src/Dotnet.Script.Tests/ScriptExecutionTests.cs
```diff
@@ -233,6 +235,17 @@ public void ShouldHandleNonExistingRemoteScript()
...
+        {
+            // This test ensures that we can load the Process class.
+            // This used to fail when executing a netcoreapp2.0 script
+            // from dotnet-script built for netcoreapp2.1
+            var result = Execute(Path.Combine("Process", "Process.csx"));
+            Assert.Contains("Success", result.output);
+        }
+
+
+        [Fact(Skip = "This also failes when run a standard netcoreapp2.1 console app")]
```
filipw: we could still run this at least on netcoreapp2.0? Something like this should do:

```
    public class OnlyOnNetCore20FactAttribute : FactAttribute
    {
        public OnlyOnNetCore20FactAttribute()
        {
                Skip = RuntimeHelper.TargetFramework != "netcoreapp2.0";
        }
    }
```

src/Dotnet.Script.Core/ScriptCompiler.cs
```diff
@@ -92,68 +83,88 @@ public virtual ScriptOptions CreateScriptOptions(ScriptContext context, IList<Ru
...
+        }
 
-            foreach (var inheritedAssemblyName in inheritedAssemblyNames)
+        private ScriptOptions AddScriptReferences(ScriptOptions scriptOptions, Dictionary<string, Assembly> loadedAssembliesMap, Dictionary<string, RuntimeAssembly> scriptDependenciesMap)
+        {
+            foreach (var runtimeAssembly in scriptDependenciesMap.Values)
             {
-                // Always prefer the resolved runtime dependency rather than the inherited assembly.
-                if(!dependencyMap.ContainsKey(inheritedAssemblyName.Name))
+                loadedAssembliesMap.TryGetValue(runtimeAssembly.Name.Name, out var loadedAssembly);
```
filipw: `runtimeAssembly.Name.Name` 😀

build/Build.csx
seesharper: Had to add the possibility to specify the target framework when doing `DotNet.Publish` since we now have multiple target frameworks. Otherwise nothing much to write home about 👍

Issue #189, opened 2017-11-27
#r is not loaded from #load .csx file


SlowLogicBoy:
Hey,

I have:
```
C:.
│   helloworld.csx
│   omnisharp.json
│
├───.vscode
│       launch.json
│
├───AnotherFolder
│       Dependencies.csx
│
└───SomeFolder
        Scripts.csx
```
Now `Dependencies.csx` contains:
```CSharp
#r "nuget: Newtonsoft.Json,10.0.3"
using Newtonsoft.Json;
Console.WriteLine(JsonConvert.SerializeObject("String"));
```
And `Scripts.csx`:
```CSharp
#load "../AnotherFolder/Dependencies.csx"
using Newtonsoft.Json;
Console.WriteLine(JsonConvert.SerializeObject("String"));
```
And `Scripts.csx` doesn't load `nuget: Newtonsoft.Json,10.0.3` unless I put nuget loading in `Scripts.csx` file. Is there a way to load references that are referenced in other `.csx` files?

```
.\New folder\AnotherFolder\Dependencies.csx(5,7): error CS0246: The type or namespace name 'Newtonsoft' could not be found (are you missing a using directive or an assembly reference?)
.\New folder\SomeFolder\Scripts.csx(6,7): error CS0246: The type or namespace name 'Newtonsoft' could not be found (are you missing a using directive or an assembly reference?)
.\New folder\AnotherFolder\Dependencies.csx(7,19): error CS0103: The name 'JsonConvert' does not exist in the current context
.\New folder\SomeFolder\Scripts.csx(8,19): error CS0103: The name 'JsonConvert' does not exist in the current context
Script compilation failed due to one or more errors.
```

EDIT:
* I noticed that if both files are in the same folder, it works.
* If I move `Scripts.csx` to the root and `Dependencies.csx` leave in `AnotherFolder` it also works.
* But if I move `Scripts.csx` to `SomeFolder`, it no longer works, I guess this is something with `../` part in  `#load "../AnotherFolder/Dependencies.csx"`
* Even if I don't use `Json.Net` in `Scripts.csx` I get the same error, since `Dependencies.csx` can't load it.
* If I don't reference any nuget packages in `Dependencies.csx` just let's say I create `SomeMethod()` it works in `Scripts.csx`

filipw:
This is a different manifestation of the same problem as mentioned here - https://github.com/filipw/dotnet-script/pull/179#discussion_r152201758.
[...]

Issue #156, opened 2017-10-30
Support inline references for script packages


filipw:
It would be great to support inline load of a package like https://www.nuget.org/packages/simple-targets-csx/. 
It's a CSX package so it contains script files that should be `#loaded`.

adamralph:
😍 Coincidentally, I started playing with using dotnet-script to execute simple-targets-csx build script this morning. This feature would make that an even more compelling use case.

filipw:
I think there is a philosophical discussion needed here, is this a `#load` or is this an `#r`? 😂

adamralph:
It's neither. It's a package reference. I'm sceptical whether #r was the
right choice for NuGet packages, but if that's how it is, then #r is the
right choice. A package could even have both assemblies and scripts.

seesharper:
I think introducing a new directive requires a lot more work than relying on existing directives such as `#load` and `#r`.  If I am not mistaken it would require changes in Roslyn to make it not complain about an unknown directive. 

The fact that we use the `#r` directive for multiple purposes sort of aligns with how we see dependencies displayed in full Visual Studio where we have Dependencies as the top level node and then we have assembly references and NuGet references below that. 

I do agree that we should use the `#r` directive here since it is actually a NuGet package and it should be relatively easy to include csx files found in the package. It would make for a great feature for dotnet-script

seesharper:
@adamralph @filipw 

I took a quick look at this and it is probably not so easy to do, but let's review our options here.

**Custom SourceResolver**

We could implement this as a custom `SourceResolver ` and provide syntax like this 

```
#load "nuget: simple-targets-csx, 5.3.0"
```

The problem here is that the `SourceResolver` is expected to return a `Stream` that points to a single file representing the resolved file. The package can contain multiple files and there is no way of knowing which is the "entry point" We could of course create syntax for specifying this like 

```
#load "nuget: simple-targets-csx/simple-targets.csx, 5.3.0"
```

that allows us to be explicit with regards to which file from the package we actually want to load.

**Combine `#r` and `#load`**

This approach splits the resolving of the NuGet package and the loading of the script into two operations.

```
#r "nuget:simple-targets-csx, 5.3.0"
#load "simple-targets.csx"
```

We would still have to implement a custom `SourceResolver` that wraps/decorates the underlying `SourceResolver` so that we can resolve the `#load` directive and load the script file from where the package gets installed which would be the global NuGet cache.

adamralph:
[...]

PR #267, opened 2018-04-26
Feature/follow loads

seesharper:
This PR adds support for following `#load` directives when resolving the set of csx files that is relevant for the current execution context.

This means that we can `#load` other scripts from anywhere on the file system and not being constrained by having all script files either in the root folder or somewhere in a folder beneath the root folder. 

A good  example of this is issue #189  which is now fixed by this PR.

*Note*
For script packages we ONLY load all scripts files when we are unable to determine the entry point script.
An entry point script is identified by a file named `main.csx`  or a single script file located at the root folder in the package. 

This aligns with the conclusion(https://github.com/filipw/dotnet-script/issues/156) we made after discussing this with the Roslyn team 

https://github.com/filipw/dotnet-script/blob/f685903a87cc5ac61fa5cb4a83615c309833f741/src/Dotnet.Script.DependencyModel/ScriptPackage/ScriptFilesDependencyResolver.cs#L31

### DIFF
--- a/src/Dotnet.Script.DependencyModel/ProjectSystem/ScriptFilesResolver.cs
+++ b/src/Dotnet.Script.DependencyModel/ProjectSystem/ScriptFilesResolver.cs
@@ -39,7 +39,7 @@ namespace Dotnet.Script.DependencyModel.ProjectSystem
 
         private static string[] GetLoadDirectives(string csxFile)
         {
-            var content = ReadFile(csxFile);
+            var content = FileUtils.ReadFile(csxFile);
             var matches = Regex.Matches(content, @"^\s*#load\s*""\s*(.+)\s*""", RegexOptions.IgnoreCase | RegexOptions.Multiline);
             List<string> result = new List<string>();
             foreach (var match in matches.Cast<Match>())
@@ -53,18 +53,6 @@ namespace Dotnet.Script.DependencyModel.ProjectSystem
             }
 
             return result.ToArray();
-        }
-
-
-        private static string ReadFile(string pathToFile)
-        {
-            using (var fileStream = new FileStream(pathToFile, FileMode.Open))
-            {
-                using (var reader = new StreamReader(fileStream))
-                {
-                    return reader.ReadToEnd();
-                }
-            }
         }
     }
 }
--- a/src/Dotnet.Script.DependencyModel/ProjectSystem/ScriptFilesResolver.cs
+++ b/src/Dotnet.Script.DependencyModel/ProjectSystem/ScriptFilesResolver.cs
@@ -1,0 +1,70 @@
+﻿using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using System.Text.RegularExpressions;
+
+namespace Dotnet.Script.DependencyModel.ProjectSystem
+{
+    public class ScriptFilesResolver
+    {
+        public HashSet<string> GetScriptFiles(string csxFile)
+        {
+            HashSet<string> result = new HashSet<string>();
+            Process(csxFile, result);
+            return result;
+        }
+
+        private void Process(string csxFile, HashSet<string> result)
+        {
+            if (result.Add(csxFile))
+            {
+                var loadDirectives = GetLoadDirectives(csxFile);
+                foreach (var loadDirective in loadDirectives)
+                {
+                    string referencedScript;
+                    if (!Path.IsPathRooted(loadDirective))
+                    {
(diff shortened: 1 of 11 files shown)


src/Dotnet.Script.DependencyModel/ProjectSystem/ScriptProjectProvider.cs
```diff
@@ -60,12 +61,24 @@ public string CreateProject(string targetDirectory, string defaultTargetFramewor
...
 
-            _logger.Debug($"Creating project file for *.csx files found in {targetDirectory} using {defaultTargetFramework} as the default framework." );
-            
+            _logger.Debug($"Creating project file for *.csx files found in {targetDirectory} using {defaultTargetFramework} as the default framework.");
+
             var csxFiles = Directory.GetFiles(targetDirectory, "*.csx", SearchOption.AllDirectories);
+            return CreateProjectFileFromScriptFiles(targetDirectory, defaultTargetFramework, csxFiles);
+        }
+
+        public string CreateProjectForScriptFile(string scriptFile)
```
filipw: can we delete the old one now? `public string CreateProject(string targetDirectory, string defaultTargetFramework = "net46", bool enableNuGetScriptReferences = false)`?

seesharper: No, that one is still used from OmniSharp where still need to look at the folder structure as a whole.
You can see that it is used from 2 different places in dotnet-script. For instance in interactive mode where there is no script file.

seesharper: https://github.com/filipw/dotnet-script/blob/033b71b22ab51768b33697d1fa4598b6c7ef0643/src/Dotnet.Script.Core/ScriptCompiler.cs#L117

src/Dotnet.Script.DependencyModel/ProjectSystem/ScriptFilesResolver.cs
```diff
@@ -0,0 +1,70 @@
...
+                result.Add(value);
+            }
+
+            return result.ToArray();
+        }
+
+
+        private static string ReadFile(string pathToFile)
+        {
+            using (var fileStream = new FileStream(pathToFile, FileMode.Open))
```
filipw: I think you'd want to do `new FileStream(filePath, FileMode.Open, FileAccess.Read, FileShare.ReadWrite)` in case files are locked

src/Dotnet.Script.DependencyModel/ProjectSystem/ScriptProjectProvider.cs
filipw: So what I would like to get rid of is this `Directory.GetFiles(targetDirectory, "*.csx", SearchOption.AllDirectories)` search.
I thin we can still do it.

 - The usage in interactive mode can be treated differently. We don't need to scan for any CSX there, we really just want to create a temporary csproj so that the base runtime assemblies for .NET Core are loaded (and then in case new nuget references are added)
 - For OmniSharp, OmniSharp itself will discover all CSX files for us, so it could just pass a list of those, and we could set projects accordingly. 

Or maybe I am missing something?
This is somewhat unrelated to this PR, but fundamentally I'd like to not have this scan all behavior.

src/Dotnet.Script.DependencyModel/ProjectSystem/ScriptFilesResolver.cs
seesharper: We do the exact same thing in the ScriptParser.
https://github.com/filipw/dotnet-script/blob/ad33be6eb98b1f69939aefa67bc2cdf4eed8a744/src/Dotnet.Script.DependencyModel/ProjectSystem/ScriptParser.cs#L106-L110

I'll extract that into one method and staying DRY

src/Dotnet.Script.DependencyModel/ProjectSystem/ScriptProjectProvider.cs
seesharper: I agree. Maybe we could do that in a separate PR? This fixes the 99% scenario where we actually execute a script located on disc. Every other scenario is left unchanged. And then we try to align the other overload with OmniSharp in a separate PR

filipw:
I think we could even leverage something like a custom script resolver now, since we pass through all the csx files, we could save the content of each and then load it from memory to improve perf

PR #269, opened 2018-04-27
issue 268 quick fix

filipw:
it's not really a solution, but it doesn't seem to break anything at least


seesharper:
All green. Merging

PR #271, opened 2018-04-29
Environmental information

seesharper:
This PR adds support for displaying environmental information that is relevant for script execution.

usage 
```
dotnet script --version
```

now yields

```
Version             : 0.23.0
Install location    : C:\Github\dotnet-script\src\Dotnet.Script\bin\Debug\netcoreapp2.0
Target framework    : netcoreapp2.0
Platform identifier : win
```

> Note : Went with extending the existing `-v|--version` option instead of `-i|--info` since that would conflict with the existing interactive mode.

This PR also cleans up the `RuntimeHelper` class making all information available as properties backed by a `Lazy<T>`

### DIFF
--- a/src/Dotnet.Script.DependencyModel/Environment/RuntimeHelper.cs
+++ b/src/Dotnet.Script.DependencyModel/Environment/RuntimeHelper.cs
@@ -79,7 +79,6 @@ namespace Dotnet.Script.DependencyModel.Environment
             return storePath;
         }
 
-
         private static string GetProcessArchitecture()
         {
             return RuntimeEnvironment.RuntimeArchitecture;            
--- a/src/Dotnet.Script.DependencyModel/Environment/RuntimeHelper.cs
+++ b/src/Dotnet.Script.DependencyModel/Environment/RuntimeHelper.cs
@@ -15,21 +15,28 @@ namespace Dotnet.Script.DependencyModel.Environment
 
         private static readonly Lazy<string> LazyInstallLocation = new Lazy<string>(GetInstallLocation);
 
-        public static string GetPlatformIdentifier()
-        {
-            if (RuntimeEnvironment.OperatingSystemPlatform == Platform.Darwin) return "osx";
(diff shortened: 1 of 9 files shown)


src/Dotnet.Script.DependencyModel/Environment/RuntimeHelper.cs
```diff
@@ -11,22 +11,33 @@ public static class RuntimeHelper
         private static readonly Regex RuntimeMatcher =
             new Regex($"{GetPlatformIdentifier()}.*-{GetProcessArchitecture()}");
 
-        private static readonly Lazy<string> LazyTargetFramework = new Lazy<string>(GetNetCoreAppVersion); 
+        private static readonly Lazy<string> LazyTargetFramework = new Lazy<string>(GetNetCoreAppVersion);
 
-        public static string GetPlatformIdentifier()
+        private static readonly Lazy<string> LazyInstallLocation = new Lazy<string>(GetInstallLocation);
```
filipw: can we package all that info into a type? I think it would make sense to have RuntimeEnvironment or something like, rather than many properties on the static helper. For now it could even be statically resolved from this helper, but would make more sense to me if things were organized that way

I can envision that in the future we will move away from the static altogether and pass around that instance

filipw:
very good - thanks

src/Dotnet.Script.DependencyModel/Environment/RuntimeHelper.cs
seesharper: This information is sort of static, but we could with little effort rename the whole `RuntimeHelper` class into `RuntimeInfo` and make it a singleton. 

The usage would then be 

```
RuntimeInfo.Instance.TargetFramework;
```

> `RuntimeInfo` is probably better so that we don't introduce confusion with the existing `Microsoft.DotNet.PlatformAbstractions.RuntimeEnvironment`

Alternatively we can call it `ScriptEnvironment`

PR #272, opened 2018-04-30
Cleanup/runtimehelper

seesharper:
This PR removes the old `RuntimeHelper` class and introduces the new `ScriptEnvironment` class that contains environmental information related to resolving dependencies, compiling and executing scripts. 

This lays down the foundation for passing/injecting `ScriptEnvironment`  in the future.

Many of the changes here are just simply static methods that now are instance methods. 
It kind of rippled through a large portion of the code base, but the overall design is now less static.


seesharper:
@filipw Are we good to merge this ? 😎

filipw:
definitely!

PR #273, opened 2018-05-01
ScriptEnvironment.IsWindows doesn't need to be internally lazy

atifaziz:
`ScriptEnvironment.IsWindows` is lazy by virtue of being a computed property that depends on the lazy property `PlatformIdentifier`.

### DIFF
--- a/src/Dotnet.Script.DependencyModel/Environment/ScriptEnvironment.cs
+++ b/src/Dotnet.Script.DependencyModel/Environment/ScriptEnvironment.cs
@@ -18,8 +18,6 @@ namespace Dotnet.Script.DependencyModel.Environment
 
         private readonly Lazy<string> _runtimeIdentifier;
 
-        private readonly Lazy<bool> _isWindows;
-
         private readonly Lazy<string> _nuGetStoreFolder;
 
         public static ScriptEnvironment Default => _default.Value;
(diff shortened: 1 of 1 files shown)


src/Dotnet.Script.DependencyModel/Environment/ScriptEnvironment.cs
```diff
@@ -30,11 +28,10 @@ private ScriptEnvironment()
             _installLocation = new Lazy<string>(GetInstallLocation);
             _platformIdentifier = new Lazy<string>(GetPlatformIdentifier);
             _runtimeIdentifier = new Lazy<string>(GetRuntimeIdentifier);
-            _isWindows = new Lazy<bool>(() => PlatformIdentifier == "win");
             _nuGetStoreFolder = new Lazy<string>(GetPathToNuGetStoreFolder);
         }
 
-        public bool IsWindows => _isWindows.Value;
+        public bool IsWindows => PlatformIdentifier == "win";
```
seesharper: Wouldn't checking the bool of an `Lazy<T>` be faster than doing string comparison every time the property is accessed?

atifaziz: Yes, that's true and you can't beat the speed of a Boolean comparison. Best one can come close to is four Boolean comparisons:

```c#
public bool IsWindows
{
    get
    {
        var pid = PlatformIdentifier;
        return pid.Length == 3 && pid[0] == 'w' && pid[1] == 'i' && pid[2] == 'n';
    }
}
```

…which is slower by 40%. I didn't realize the purpose of making `IsWindows` lazy internally was micro-optimization. [`RuntimeEnvironment.OperatingSystemPlatform` is internally lazy](https://github.com/dotnet/core-setup/blob/9dff3ce12574699e6d4d31722047f775d92d05fd/src/managed/Microsoft.DotNet.PlatformAbstractions/Native/PlatformApis.cs#L18), then so is `ScriptEnvironment.PlatformIdentifier` and `ScriptEnvironment.IsWindows` on top of that, which seemed a bit excessive. Generally, if people are concerned about the performance of a property access/computation in a very tight loop then they'll cache the property access in a local variable on the stack.

Happy if you wish to keep things as-are and close this issue.

seesharper:
> Happy if you wish to keep things as-are and close this issue.

Decided to keep it as is for this one 👍

PR #274, opened 2018-05-01
ScriptEnvironment.Default laziness is redundant

atifaziz:
The `ScriptEnvironment.Default` doesn't need to be made lazy as class initialization happens lazily anyway. According to the C# language specification:

> #### [Static field initialization](https://github.com/dotnet/csharplang/blob/0760caef1b0e81e0ea0d869dbda9c6e411dc7055/spec/classes.md#static-field-initialization)
>
> The static field variable initializers of a class correspond to a sequence of assignments that are executed in the textual order in which they appear in the class declaration. If a static constructor (Static constructors) exists in the class, execution of the static field initializers occurs immediately prior to executing that static constructor. Otherwise, the static field initializers are executed at an implementation-dependent time prior to the first use of a static field of that class.


atifaziz:
I'm aware of those articles but even [he concluded](http://csharpindepth.com/Articles/General/Singleton.aspx#conclusion):

> My personal preference is for [solution 4](http://csharpindepth.com/Articles/General/Singleton.aspx#cctor): the only time I would normally go away from it is if I needed to be able to call other static methods without triggering initialization, or if I needed to know whether or not the singleton has already been instantiated. I don't remember the last time I was in that situation, assuming I even have.

The following used to be the case:

> additional "trickery" to ensure laziness and correctness.

But [type initialization has been quite lazy since CLR 4](https://codeblog.jonskeet.uk/2010/01/26/type-initialization-changes-in-net-4-0/) so we can all put that sorcery behind us. :wink:

> It is also more readable to the novice in that it expresses the intent more clearly.

Actually, some novice might say, “Yuck! This is C#?” 😆
 
I found the code to be “paranoid” when reading it so the idea with this PR was to suggest removal of redundant ceremony. That's all. If you wish to keep it the way you have it now than I am happy to close the PR.

seesharper:
> But type initialization has been quite lazy since CLR 4 so we can all put that sorcery behind us. 

I was not aware of that 👍 

And with that out of the way, I'll agree this looks cleaner. Merging

PR #277, opened 2018-05-01
Upping Docker build-baseimage to 2.1-sdk

andmos:
Restore failed with the `2.0.0-sdk`, so upping to `dotnet:2.1-sdk` for the builder.


build/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM microsoft/dotnet:2.0.0-sdk as builder
+FROM microsoft/dotnet:2.1-sdk as builder
```
seesharper: LGTM, this is because we cross compile for both `netcoreapp2.0` and `netcoreapp2.1` now 👍

andmos: Yeah, saw that while building. Great to do in the Dockerfile as well!

PR #278, opened 2018-05-02
Support using Dotnet.Script.Core on Desktop CLR

filipw:
This is a first step for #44 

The core library is netstandard2.0 so users can install it into i.e. net461 projects. There is no reason why we'd prevent them from doing so or why we shouldn't support that scenario.

In principle what it means is that we should detect the version of .NET. This is not an easy thing to do and is often approximation so for now my suggested solution is to expose something like `OverrideTargetFramework("net461")` which you can statically call (in the future, we'll move to passing around an instance of `ScriptEnvironment` and it will be a bit more elegant, without global state). This, by coincidence, allows you to also enforce i.e. `netcoreapp2.0` resolution mode from `netcoreapp2.1` host.
When the target framework is not .NET Core, we can still use the new project system and the assets file, but the base references like `System`, `mscorlib`, `System.Core` or `System.Runtime` have to be added implicitly. In this case it's assumed they are safe to be inherited from the host.

In the next step, we should introduce a desktop CLR runner for the command line.

We have a new test assembly now for Desktop tests. I currently only set up the REPL tests to run there, because they run in proc so are easy to execute. Once we have a command line runner we could run the rest of the tests too. The desktop tests are configured to run for Windows builds only atm, mainly because `dotnet test` doesn't support non-.NET Core test projects on Unix. The more sustainable solutions for long term would be to use xunit command line runner on Mono.


seesharper:
Of course 😄

filipw:
grün

build/BuildContext.csx
```diff
@@ -3,75 +3,35 @@ using static FileUtils;
...
-
-PathToReleaseNotes = Path.Combine(GitHubArtifactsFolder, "ReleaseNotes.md");
-
-Version = ReadVersion();
-
-PathToGitHubReleaseAsset = Path.Combine(GitHubArtifactsFolder, $"{ProjectName}.{Version}.zip");
+var owner = "filipw";
+var projectName = "dotnet-script";
+var root = FileUtils.GetScriptFolder();
+var solutionFolder = Path.Combine(root,"..","src");
```
seesharper: Nice cleanup 👍

PR #280, opened 2018-05-03
Added nuget badge

filipw:
So that latest version is shown in readme

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,13 +1,13 @@
 # dotnet script
 
-Run C# scripts from the .NET CLI.
+Run C# scripts from the .NET CLI.  [![Nuget](http://img.shields.io/nuget/v/dotnet-script.svg?maxAge=10800)](https://www.nuget.org/packages/dotnet-script/)
 
 ## Build status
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,13 +1,13 @@
 # dotnet script
 
-Run C# scripts from the .NET CLI.
+Run C# scripts from the .NET CLI.  [![Nuget](http://img.shields.io/nuget/v/dotnet-script.svg?maxAge=10800)](https://www.nuget.org/packages/dotnet-script/)
```
seesharper: Would it make sense to list the other packages as well?
* Dotnet.Script.Core
* Dotnet.Script.DependencyModel

Maybe formatted in a table like the build status?

seesharper: The `Dotnet.Script` package is soon to be deprecated and the `Dotnet.Script.DependencyModel.NuGet` is sort of just a transitive so these does not need to be listed

filipw: I added all for now, after all ,they are public 😄 
once we deprecate one of them, I think it's even better to mention it here so that it's not confusing when someone searches nuget

PR #282, opened 2018-05-04
Feature/info option

seesharper:
This PR adds support for the `--info` option that displays environmental information related to resolving dependencies, compiling and executing scripts.

```
dotnet script -v|--version
```

yields

```
0.23.0
```

while 

```
dotnet script --info
```

yields 

```
Version             : 0.23.0
Install location    : C:\Github\dotnet-script\src\Dotnet.Script\bin\Debug\netcoreapp2.0
Target framework    : netcoreapp2.0
Platform identifier : win
```

Also added the `ScriptTestRunner` class as a start to clean up much of the stuff going on in tests.
There is currently a lot of duplication of code related to script arguments and such. 
Will do an overall cleanup in a separate PR 👍

### DIFF
--- a/src/Dotnet.Script.Tests/EnvironmentTests.cs
+++ b/src/Dotnet.Script.Tests/EnvironmentTests.cs
@@ -1,0 +1,37 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+using Xunit;
+using System.Text.RegularExpressions;
+
+
+namespace Dotnet.Script.Tests
+{
+    public class EnvironmentTests
+    {
+        [Fact]
+        public void ShouldPrintVersionNumber()
+        {
+            var result = ScriptTestRunner.Default.Execute("--version");
+            Assert.Equal(0, result.exitCode);
+            Assert.Matches(@"\d*.\d*.\d*", result.output);
+
+            result = ScriptTestRunner.Default.Execute("-v");
+            Assert.Equal(0, result.exitCode);
+            Assert.Matches(@"\d*.\d*.\d*", result.output);
+        }
+
+        [Fact]
+        public void ShouldPrintInfo()
+        {
+            var result = ScriptTestRunner.Default.Execute("--info");
+            Assert.Equal(0, result.exitCode);
+            Assert.Contains("Version", result.output);
+            Assert.Contains("Install location", result.output);
+            Assert.Contains("Target framework", result.output);
+            Assert.Contains("Platform identifier", result.output);
+            Assert.Contains("Runtime identifier", result.output);
+        }
+
+    }
+}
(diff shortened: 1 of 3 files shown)


filipw:
Thanks!

src/Dotnet.Script.Tests/ScriptTestRunner.cs
```diff
@@ -0,0 +1,66 @@
...
+        }
+
+        public int ExecuteInProcess(params string[] arguments)
+        {                        
+            return Program.Main(arguments ?? Array.Empty<string>());
+        }
+
+        public (string output, int exitCode) ExecuteFixture(string fixture, params string[] arguments)
+        {
+            var pathToFixture = Path.Combine("..", "..", "..", "TestFixtures", fixture);
```
atifaziz: Reuse `GetPathToFixture` here?

src/Dotnet.Script.Tests/ScriptTestRunner.cs
```diff
@@ -0,0 +1,66 @@
...
+            var result = ProcessHelper.RunAndCaptureOutput("dotnet", GetDotnetScriptArguments(new string[] { pathToFixture}.Concat(arguments ?? Array.Empty<string>()).ToArray()));
+            return result;
+        }
+
+        public int ExecuteFixtureInProcess(string fixture, params string[] arguments)
+        {
+            var pathToFixture = Path.Combine("..", "..", "..", "TestFixtures", fixture);
```
atifaziz: This is not used. Can be removed as you call `GetPathToFixture` below.

src/Dotnet.Script.Tests/EnvironmentTests.cs
```diff
@@ -0,0 +1,37 @@
...
+namespace Dotnet.Script.Tests
+{
+    public class EnvironmentTests
+    {
+        [Fact]
+        public void ShouldPrintVersionNumber()
+        {
+            var result = ScriptTestRunner.Default.Execute("--version");
+            Assert.Equal(0, result.exitCode);
+            Assert.Matches(@"\d*.\d*.\d*", result.output);
```
atifaziz: This is an incorrect test. It says match 3 _Unicode digit_ separated by _anything_ and which appears _anywhere_ in the output. What's more, since the Unicode digits are _optional_ (`*`), it will also match 3 of anything! The right test would be:

```c#
Assert.Matches(@"^[0-9]+(\.[0-9]+){2}$", result.output);
```

However, this test is still somewhat inaccurate. If you run `dotnet script --version`, the output is:

```

0.23.0
```

The trouble here is that there is a leading blank line! The test shouldn't be passing but looks like it does because `ProcessHelper.RunAndCaptureOutput` trims the output:

https://github.com/filipw/dotnet-script/blob/9bced12676415fd5eb1c69b95cf3765c71190358/src/Dotnet.Script.Tests/ProcessHelper.cs#L40

Ideally trimming should be decision of the caller.

atifaziz:
Never mind [my review](https://github.com/filipw/dotnet-script/pull/282#pullrequestreview-117618986), [this got merged](https://github.com/filipw/dotnet-script/commit/9bced12676415fd5eb1c69b95cf3765c71190358) while I writing up. 😛

PR #283, opened 2018-05-04
Fix version test regex pattern

atifaziz:
This PR applied fixes suggested in [my (later) review](https://github.com/filipw/dotnet-script/pull/282#pullrequestreview-117618986) of #282.


filipw:
thanks!

PR #287, opened 2018-05-08
Avoid loading version info unless option is specified

atifaziz:
This should also improve start-up performance a teeny-weeny bit because the version information was being loaded on each and every run irrespective of whether the version option (`-v` or `--version`) was specified.


filipw:
good catch - thanks!

PR #295, opened 2018-05-16
Add .csx extension to a new file if not specified

SlowLogicBoy:
This is pretty annoying -.-


filipw:
thanks, can you add a test here https://github.com/filipw/dotnet-script/blob/master/src/Dotnet.Script.Tests/ScaffoldingTests.cs#L72 for this use case?

SlowLogicBoy:
On the same test? shouldn't I create 2 new tests: one for `init` and one for `new`?

SlowLogicBoy:
Right now I just added additional lines to existing tests, waiting for your response.

filipw:
oops sorry for not being clear - it would make sense to have a different tests for this feature 😀
this way when they fail, it's explicit what has failed, instead of having to dig through asserts

filipw:
thanks!

PR #310, opened 2018-06-07
Feature/nuget sources

seesharper:
This PR adds support for specifying package sources when executing a script.

The syntax is identical to the syntax used by the `dotnet cli`

```
dotnet script foo.csx -s http://SomePackageSource
```

It is implemented simply by forwarding the package sources to the `dotnet restore` command we use to resolve runtime and compilation dependencies. 

This change kind of rippled through the code base as we need to pass these package sources down the line. 

Also added some logging to the `CommandRunner`class so that we log the command and the arguments used to execute the command.


filipw:
ooooh 💄✨

src/Dotnet.Script.Core/ScriptContext.cs
```diff
@@ -9,14 +10,15 @@ namespace Dotnet.Script.Core
 {
     public class ScriptContext
     {
-        public ScriptContext(SourceText code, string workingDirectory, IEnumerable<string> args, string filePath = null, OptimizationLevel optimizationLevel = OptimizationLevel.Debug, ScriptMode scriptMode = ScriptMode.Script)
+        public ScriptContext(SourceText code, string workingDirectory, IEnumerable<string> args, string filePath = null, OptimizationLevel optimizationLevel = OptimizationLevel.Debug, ScriptMode scriptMode = ScriptMode.Script, string[] packageSources = null)
```
filipw: these constructors are starting to resemble some of the constructors in Roslyn 😀

filipw:
> It is implemented simply by forwarding the package sources to the dotnet restore command we use to resolve runtime and compilation dependencies.

so not written into csproj? I wonder why we didn't think of that earlier

seesharper:
I actually started to put them into the project file when I suddenly realized that we could just offload this directly to the dotnet cli 😎

seesharper:
I see the build failed. I’ll look into it tomorrow 👍

seesharper:
@filipw Finally green across the board

Issue #227, opened 2018-02-25
Add support for outputting into a DLL


filipw:
I'd like to be able to compile a script into a DLL for later use.

This feature would probably need two parts:

 - *add ability to compile into a DLL.* The usage of such a DLL would however be somewhat limited, as you'd need to be familiar with the script code layout once the assembly is emitted (the wrapper submission class etc). We may want to give the submission wrapper class a better name for better discoverability
 - *add ability to run from a DLL.* Given the peculiar entry point into scripting (the `<Factory>` method), we could add some code that simplifies running a compiled DLL too

seesharper:
Just out of curiosity, what is the use case here? 😄

SlowLogicBoy:
Executable would be nice too (dotnet publish -r <yourplatform>), since there been quite a few time where I needed to make an executable from .csx file.

filipw:
I think the exe would be a sexy feature, but it's a lot more complicated than the others - we could add it to the list though, makes sense.

> Just out of curiosity, what is the use case here?

startup time (for example FAKE compiles to DLL to improve subsequent execution times), distribute to others without possibility to edit, work with scenarios where in-memory assembly is not enough (i.e. plugin discovery)

PR #312, opened 2018-06-09
Publish a .csx script as a binary executable

Sharpiro:
This PR allows for publishing a script as an executable.

I found myself every now and then having a good use case to basically take some scripts I've created and just have them outputted to an executable file.  I thought this would be kind of cool and looked relevant to the current issue: #227.  However this is slightly different in that this isn't currently concerned with outputting scripts to re-usable libraries, and just executes as if ```dotnet script main.csx``` were called, except from a .exe file.  If this isn't something you guys want in the main code base, I understand.  I also did my best to modify as few files as possible and follow any coding styles I noticed while learning the code base.

I created a new "publish" command that takes in a script argument, an optional publish directory, and a debug flag.  It uses a lot of code that you guys have already written, but goes a bit further with modifying the generated temp project, adding a Program.cs template, generating an assembly from the scripts, and moving around any referenced .dll files as needed.  ```dotnet publish``` is then executed from your command runner class and this creates the executable.

Let me know if you have any questions or if there is an obvious better way to do this.  Thanks.

Example usage:
```
dotnet script publish C:\Users\sharpiro\Desktop\temp\tempScript\main.csx -d -o C:\Users\sharpiro\Desktop\temp\publishTemp
```


filipw:
Thanks a lot, this is a great idea, and this is definitely a feature that we want.
At the high level, this implementation is also along the lines of what we'd want to have (compile, reference the DLL and run precompiled code using `<Factory>` method). 

You can have a look at my PoC for #227 which is really very similar, it's here https://github.com/filipw/dotnet-script/commit/e677cd63fd3bfd22ebbbf026b82b0a2167a1c447

I think what I'd like to see is if you incorporate the `ScriptEmitter` from that commit; this would decouple the process of emitting from the publish, which will be helpful in the future.

Can you also add some tests? I'll also add some comments inline - thanks!

src/Dotnet.Script.Core/ScriptPublisher.cs
```diff
@@ -0,0 +1,113 @@
...
+        private string CreateScriptAssembly(ScriptContext context, string tempProjectDirecory)
+        {
+            try
+            {
+                var compilationContext = _scriptCompiler.CreateCompilationContext<int, CommandLineScriptGlobals>(context);
+                var scriptOptions = compilationContext.Script.GetCompilation().Options
+                    .WithScriptClassName(AssemblyName);
+                var scriptCompilation = compilationContext.Script.GetCompilation()
+                      .WithOptions(scriptOptions)
+                    .WithAssemblyName(AssemblyName);
```
filipw: is the assembly name reset really necessary?

src/Dotnet.Script.Core/ScriptPublisher.cs
```diff
@@ -0,0 +1,113 @@
...
+        public void CreateExecutable(ScriptContext context, LogFactory logFactory)
+        {
+            var tempProjectPath = _scriptProjectProvider.CreateProjectForScriptFile(context.FilePath);
+            var tempProjectDirecory = Path.GetDirectoryName(tempProjectPath);
+
+            var scriptAssemblyPath = CreateScriptAssembly(context, tempProjectDirecory);
+
+            var projectFile = new ProjectFile(File.ReadAllText(tempProjectPath));
+            // todo: grab version in a better way?
+            projectFile.AddPackageReference(new PackageReference("Microsoft.CodeAnalysis.Scripting", "2.8.2", PackageOrigin.ReferenceDirective));
```
filipw: I think it's ok to hardcode a version. You could technically omit it, with .net core SDK 2.1.300 it would just pull the latest but it might conflict with something. Let's make a global constant for the version

src/Dotnet.Script.Core/ScriptPublisher.cs
```diff
@@ -0,0 +1,113 @@
...
+                var compilationContext = _scriptCompiler.CreateCompilationContext<int, CommandLineScriptGlobals>(context);
+                var scriptOptions = compilationContext.Script.GetCompilation().Options
+                    .WithScriptClassName(AssemblyName);
+                var scriptCompilation = compilationContext.Script.GetCompilation()
+                      .WithOptions(scriptOptions)
+                    .WithAssemblyName(AssemblyName);
+
+                foreach (var reference in scriptCompilation.DirectiveReferences)
+                {
+                    var refInfo = new FileInfo(reference.Display);
```
filipw: you may want to check for `nuget` pragma here, if it's a nuget reference there is nothing to copy (it resolves to a fake DLL) so you can skip it.

src/Dotnet.Script.Core/ScriptPublisher.cs
```diff
@@ -0,0 +1,113 @@
...
+                    File.Copy(refInfo.FullName, $"{tempProjectDirecory}/{refInfo.Name}", true);
+                }
+
+                var assemblyPath = $"{tempProjectDirecory}/{AssemblyName}.dll";
+                var emitResult = scriptCompilation.Emit(assemblyPath);
```
filipw: let's incorporate `ScriptEmitter` instead https://github.com/filipw/dotnet-script/commit/e677cd63fd3bfd22ebbbf026b82b0a2167a1c447

src/Dotnet.Script.Core/Templates/program.publish.template
```diff
@@ -0,0 +1,50 @@
...
+            try
+            {
+                var globals = new CommandLineScriptGlobals(Console.Out, CSharpObjectFormatter.Instance);
+                foreach (var arg in args)
+                    globals.Args.Add(arg);
+
+                var factoryMethod = typeof(scriptAssembly).GetMethod("<Factory>");
+                if (factoryMethod == null) throw new Exception("couldn't find factory method to initiate script");
+
+                // todo: not sure what the second parameter is, using 'null' for now
```
filipw: this is OK - the second index in the array is for submission continuations (REPL scenarios)

src/Dotnet.Script.Core/Templates/program.publish.template
```diff
@@ -0,0 +1,50 @@
...
+using Microsoft.CodeAnalysis.CSharp.Scripting.Hosting;
+using Microsoft.CodeAnalysis.Scripting.Hosting;
+using System.Threading.Tasks;
+using static System.Console;
+
+namespace dotnetPublishCode
+{
+    class Program
+    {
+        static void Main(string[] args)
```
filipw: you can change it to `async Main` and then there is no need to block in the code

src/Dotnet.Script.Core/Templates/program.publish.template
```diff
@@ -0,0 +1,50 @@
...
+
+                var factoryMethod = typeof(scriptAssembly).GetMethod("<Factory>");
+                if (factoryMethod == null) throw new Exception("couldn't find factory method to initiate script");
+
+                // todo: not sure what the second parameter is, using 'null' for now
+                var invokeTask = factoryMethod.Invoke(null, new object[] { new object[] { globals, null } }) as Task<int>;
+                var invokeResult = invokeTask.Result;
+                if (invokeResult != 0) WritePrettyError($"Error result: '{invokeResult}'");
+            }
+            // todo: not getting the full script stack trace
```
filipw: what do you mean?

src/Dotnet.Script.Core/Templates/program.publish.template
```diff
@@ -0,0 +1,50 @@
...
+            catch (AggregateException ex)
```
filipw: I think you want something like this

```
                if (e is AggregateException aggregateEx)
                {
                    e = aggregateEx.Flatten().InnerException;
                }
```

src/Dotnet.Script.Core/ScriptPublisher.cs
Sharpiro: k I'll add something for this, I noticed what you're talking about

src/Dotnet.Script.Core/Templates/program.publish.template
Sharpiro: so i noticed that when i would throw an error in the script, I wouldn't get any of the script stack trace information to print out.  I can see if that was because I wasn't using your emitter with debug information maybe?

src/Dotnet.Script.Core/Templates/program.publish.template
Sharpiro: wouldn't i need to still loop through the flattened "InnerExceptions" of a flattened aggregate exception typically?  I think you're right though in this case there will only ever be 1 inner exception in that aggregate exception.

src/Dotnet.Script.Core/ScriptPublisher.cs
Sharpiro: I'll check on this and make sure

Sharpiro:
Thanks for the feedback.  I'll try to update this code to use your script emitter, make adjustments based on your comments, and add a few tests hopefully tomorrow.

src/Dotnet.Script.Core/Templates/program.publish.template
Sharpiro: for example, if i threw an error on line 25 of the main.csx file, when writing out the error it wouldn't give me any indication that the error actually occurred on 25, or if there was previous stack trace data

filipw: this is probably because you emit without symbols - if you compare the emitter code I mentioned before, it emits symbols too and line numbers are available in the symbols only

src/Dotnet.Script.Core/Templates/program.publish.template
filipw: I think on these code paths there can only ever be one

Sharpiro:
I just committed code based on your feedback.

I found that I did indeed need to modify the assembly name since prior to that it was an unspeakable name and would also be an invalid file name.  Even when the assembly name was kept the same, but just written to a valid file name, there would be an error where the code would be looking for that unspeakable assembly name. 

I'll go ahead and start working on some tests soon.

Sharpiro:
So how does error suppression work here?  I noticed you're suppressing CS1705 which is an error, not a warning like the others. Even though you are supprsssing this error when evaluating the script diagnostics, won't this error just cause the script or emit to fail later on anyways?  I tested a different error and tried to suppress it through your process, but it just causes the emit to fail in the script emitter later on. 

Also I don't see any current reason to suppress warnings. From what I can see dotnet script has no functionality or perhaps no desire to log or print warnings to the user.  It simply logs the diagnostics it suppressed and then errors and logs any unsuppressed errors if any were encountered.

seesharper:
The need for suppressing CS1705 actually went away with https://github.com/filipw/dotnet-script/pull/261. We decided to keep it nevertheless in case we still had assembly loading issues. 
We are probably okay not suppressing it 👍

src/Dotnet.Script.Core/ScriptEmitResult.cs
```diff
@@ -0,0 +1,35 @@
...
+            PeStream = peStream;
+            PdbStream = pdbStream;
+            DirectiveReferences = directiveReferences.ToImmutableArray();
+        }
+
+        public MemoryStream PeStream { get; }
+        public MemoryStream PdbStream { get; }
+        public ImmutableArray<Diagnostic> Diagnostics { get; private set; } = ImmutableArray.Create<Diagnostic>();
+        public ImmutableArray<MetadataReference> DirectiveReferences { get; } = ImmutableArray.Create<MetadataReference>();
+        public bool IsErrored => Diagnostics.Any();
```
seesharper: `Success` might be a better name her since that aligns with `EmitResult.Success`

src/Dotnet.Script.Core/ScriptPublisher.cs
```diff
@@ -0,0 +1,121 @@
...
+        {
+            try
+            {
+                var emitResult = _scriptEmitter.Emit<int>(context, AssemblyName);
+                if (emitResult.IsErrored)
+                {
+                    throw new CompilationErrorException("One or more errors occurred when emitting the assembly", emitResult.Diagnostics);
+                }
+
+                var assemblyPath = $"{tempProjectDirecory}/{AssemblyName}.dll";
```
seesharper: Use Path.Combine to ensure correctness across all operating systems 👍

seesharper:
Left a few nit picking comments. Otherwise LGTM. Oh and the branch needs updating 😄

filipw:
> I noticed you're suppressing CS1705

this has been historically suppressed in the .NET CLI as well as in scripting, because the runtime should redirect to the correct version. 
OmniSharp suppresses those too https://github.com/OmniSharp/omnisharp-roslyn/blob/76810f8655fbd198187b8ef0eae8bd6a3bee3c4c/src/OmniSharp.Roslyn/Utilities/CompilationOptionsHelper.cs#L10-L16

src/Dotnet.Script.Core/ScriptEmitResult.cs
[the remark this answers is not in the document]
filipw: also, diagnostics would also contains warnings. I think you should just check if there are any diagnostics that are error level

seesharper:
@Sharpiro This all looks good although I would love to see some basic testing before merging 😎

Sharpiro:
just added a few simple tests and am now using absolute file paths in the same way the default app command is.

Sharpiro:
Looks like the Travis CI build didn't like [my tests](https://github.com/Sharpiro/dotnet-script/blob/a517c4e9f39bad4c521616ddd487171f147f4b77/src/Dotnet.Script.Tests/ScriptPublisherTests.cs)

The tests work on my machine as well as AppVeyor.  Anything I need to know about Travis CI?

seesharper:
Okay, a couple of things here. When compiling a self contained binary on *nix, they don't get the `.exe` extension.  Also I would like to have a test that actually executes the generated binary 👍

seesharper:
Now we are in business 🎸

PR #315, opened 2018-06-16
Feature/create and execute dll

Sharpiro:
This feature continues with my previous publish to .exe PR #312  as well as Filip's POC e677cd6 to create and execute a .dll .

I made improvements to the way we were emitting the streams in ScriptEmitter:
- Debug information is now embedded in the PeStream so we have 1 less file to worry about (and apparently I hear it might be more performant?).  
- If the publish option for ```-c release``` is supplied, then no debug information will be created.  
- It tuns out Filip was right and I did not need to actually rename the assembly to get the publishing working, so I removed that bit of code.

.dll generation considerations:
- when the ```--dll``` flag is passed to the ```publish``` command, the script is outputted to a .dll in a publish directory in the working directory (configurable), along w/ any .dll's the script referenced via ```#r```.  Getting Nuget packages working was trickier.  I elected to not move over the assembly files to the publish directory during a publish, instead moving 2 files from the "obj" folder, "project.assets.json", and "script.csproj.nuget.g.props" to determine the necessary nuget information when the generated .dll is executed.  1 caveat to doing this, is no nuget restore takes place at execution time.  I figured this would be ok after I tested the ```dotnet test.dll``` command and saw that it does not restore nuget packages either.  This does mean though that if you do a publish to dll, delete your nuget packages (or execute from a machine missing the packages), then try to execute the dll, you will get a missing reference error.
- A dll is generated and named based upon the current working directory.  This was not possible to do with .exe generation w/o some breaking changes.
- After a dll has been generated, it can then be run through dotnet script via the ```exec``` command.  This command calls the ScriptRunner, creates the dependency map from the obj information, runs the main script ```<Factory>``` method, and then resolves missing nuget references.

misc:
- it seems d93a5572 may have broken the ```Dotnet.Script.Desktop.Tests``` tests

Looking forward to your feedback.  Thanks

### DIFF
--- a/src/Dotnet.Script.Core/ScriptPublisher.cs
+++ b/src/Dotnet.Script.Core/ScriptPublisher.cs
@@ -12,8 +12,7 @@ namespace Dotnet.Script.Core
 {
     public class ScriptPublisher
     {
-        const string AssemblyName = "scriptAssembly";
-        const string ScriptingVersion = "2.8.2";
+        private const string ScriptingVersion = "2.8.2";
 
         private readonly ScriptProjectProvider _scriptProjectProvider;
         private readonly ScriptEmitter _scriptEmitter;
@@ -38,16 +37,16 @@ namespace Dotnet.Script.Core
         {
         }
 
-        public void CreateAssembly(ScriptContext context, LogFactory logFactory)
+        public void CreateAssembly<TReturn, THost>(ScriptContext context, LogFactory logFactory, string assemblyFileName = null)
         {
             Directory.CreateDirectory(context.WorkingDirectory);
             Directory.CreateDirectory(Path.Combine(context.WorkingDirectory, "obj"));
 
+            assemblyFileName = assemblyFileName ?? Path.GetFileNameWithoutExtension(context.FilePath);
+            CreateScriptAssembly<TReturn, THost>(context, context.WorkingDirectory, assemblyFileName);
+
             var tempProjectPath = ScriptProjectProvider.GetPathToProjectFile(Path.GetDirectoryName(context.FilePath));
             var tempProjectDirecory = Path.GetDirectoryName(tempProjectPath);
-            var assemblyFileName = Path.GetFileName(tempProjectDirecory);
-            CreateScriptAssembly(context, context.WorkingDirectory, assemblyFileName);
-
 
             var sourceProjectAssetsPath = Path.Combine(tempProjectDirecory, "obj", "project.assets.json");
             var destinationProjectAssetsPath = Path.Combine(context.WorkingDirectory, "obj", "project.assets.json");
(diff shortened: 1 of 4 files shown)


src/Dotnet.Script/Program.cs
```diff
@@ -133,17 +133,20 @@ private static int Wain(string[] args)
                 c.Description = "Creates an executable from a script";
```
filipw: please update the description here

src/Dotnet.Script/Program.cs
```diff
@@ -133,17 +133,20 @@ private static int Wain(string[] args)
...
                 var fileNameArgument = c.Argument("filename", "The script file name");
                 var publishDirectoryOption = c.Option("-o |--output", "Directory where the published executable should be placed.  Defaults to a 'publish' folder in the current directory.", CommandOptionType.SingleValue);
+                var dllOption = c.Option("--dll", "Publish to a .dll instead of a .exe", CommandOptionType.NoValue);
+                var commandConfig = c.Option("-c | --configuration <configuration>", "Configuration to use for running the script [Release/Debug] Default is \"Debug\"", CommandOptionType.SingleValue);
                 var publishDebugMode = c.Option(DebugFlagShort + " | " + DebugFlagLong, "Enables debug output.", CommandOptionType.NoValue);
                 c.OnExecute(() =>
                 {
+                    var x = debugMode.HasValue();
```
filipw: what's `x`? 😀

src/Dotnet.Script/Program.cs
```diff
@@ -133,17 +133,20 @@ private static int Wain(string[] args)
                 c.Description = "Creates an executable from a script";
                 var fileNameArgument = c.Argument("filename", "The script file name");
                 var publishDirectoryOption = c.Option("-o |--output", "Directory where the published executable should be placed.  Defaults to a 'publish' folder in the current directory.", CommandOptionType.SingleValue);
+                var dllOption = c.Option("--dll", "Publish to a .dll instead of a .exe", CommandOptionType.NoValue);
+                var commandConfig = c.Option("-c | --configuration <configuration>", "Configuration to use for running the script [Release/Debug] Default is \"Debug\"", CommandOptionType.SingleValue);
```
filipw: why do we need an extra option instead of the already existing top-level ` var configuration = app.Option("-c | --configuration <configuration>"`?

src/Dotnet.Script/Program.cs
Sharpiro: I hate that visual studio didn't give a warning for this type of unused variable...

src/Dotnet.Script/Program.cs
Sharpiro: I'm glad you mentioned this. I tested trying to access "app" level "options" when executing a sub "command", and those options were not coming through and required me to create new variables in order to access them at the "command" level.

Sharpiro:
So for the generated DLL name, since the name of the script is usually less descriptive than the folder, I went with the folder. Using ```dotnet script init``` the intro script is always just "main.csx", whereas the folder might describe the script project. I don't have a strong opinion and can change this to just the script name if you'd like.

src/Dotnet.Script/Program.cs
filipw: I am barely able to handle this argument parsing library 😀 so I now suspect this is happening because the `publish` is a command so it creates a new context and all the arguments are parsed within this context, not in the global one?

filipw:
I think we should use the script name. Perhaps we can have another switch, something like `-o | --output` where we can override the name?

src/Dotnet.Script.Core/ScriptPublisher.cs
```diff
@@ -60,35 +80,28 @@ public void CreateExecutable(ScriptContext context, LogFactory logFactory)
             if (exitcode != 0) throw new Exception($"dotnet publish failed with result '{exitcode}'");
         }
 
-        private string CreateScriptAssembly(ScriptContext context, string tempProjectDirecory)
+        private string CreateScriptAssembly(ScriptContext context, string outputDirectory, string assemblyFileName)
         {
             try
             {
-                var emitResult = _scriptEmitter.Emit<int>(context, AssemblyName);
+                var emitResult = _scriptEmitter.Emit<int>(context);
```
filipw: I think we can remove the `PdbStream` property from the emitResult?

src/Dotnet.Script/Program.cs
Sharpiro: Yeah that has been the assumption I went with as well.

src/Dotnet.Script.Core/ScriptPublisher.cs
Sharpiro: Yea I already removed it since the debug information is embedded

PR #321, opened 2018-06-22
 Added support for passing in runtime ID when doing publish to exe

filipw:
When publishing to an exe you can now pass the runtime ID to create a self contained executable for a different platform than your current one.
For example: `dotnet script publish main.csx --runtime osx-x64`. 

If the runtime is not passed, the current runtime is used (as it was the case earlier). 
If the output directory is passed, the output directory is used.
If no output directory is passed, the default path is computed the following way `{current dir}/publish/{runtime id}`.

cc @Sharpiro

### DIFF
--- a/src/Dotnet.Script/Program.cs
+++ b/src/Dotnet.Script/Program.cs
@@ -131,14 +131,14 @@ namespace Dotnet.Script
 
             app.Command("publish", c =>
             {
-                c.Description = "Creates an executable or DLL from a script";
+                c.Description = "Creates a self contained executable or DLL from a script";
                 var fileNameArgument = c.Argument("filename", "The script file name");
                 var publishDirectoryOption = c.Option("-o |--output", "Directory where the published executable should be placed.  Defaults to a 'publish' folder in the current directory.", CommandOptionType.SingleValue);
-                var dllName = c.Option("-n |--name", "The name for the generated DLL (EXE not supported at this time).  Defaults to the name of the script.", CommandOptionType.SingleValue);
-                var dllOption = c.Option("--dll", "Publish to a .dll instead of a .exe", CommandOptionType.NoValue);
-                var commandConfig = c.Option("-c | --configuration <configuration>", "Configuration to use for running the script [Release/Debug] Default is \"Debug\"", CommandOptionType.SingleValue);
+                var dllName = c.Option("-n |--name", "The name for the generated DLL (executable not supported at this time).  Defaults to the name of the script.", CommandOptionType.SingleValue);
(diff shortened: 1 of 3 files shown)


src/Dotnet.Script/Program.cs
```diff
@@ -138,6 +138,8 @@ private static int Wain(string[] args)
                 var dllOption = c.Option("--dll", "Publish to a .dll instead of a .exe", CommandOptionType.NoValue);
                 var commandConfig = c.Option("-c | --configuration <configuration>", "Configuration to use for running the script [Release/Debug] Default is \"Debug\"", CommandOptionType.SingleValue);
                 var publishDebugMode = c.Option(DebugFlagShort + " | " + DebugFlagLong, "Enables debug output.", CommandOptionType.NoValue);
+                var runtime = c.Option("-r |--runtime", "The runtime to publish the self contained EXE to. Defaults to the your current platform.", CommandOptionType.SingleValue);
```
seesharper: Should maybe be "Defaults to your current runtime" instead of `platform`. So we don't mix up the terminology :)
Also maybe "The runtime to publish the self contained executable to" instead of "EXE" since that is sort of Windows only :)

filipw: makes sense, I updated the texts

src/Dotnet.Script/Program.cs
```diff
@@ -131,14 +131,14 @@ private static int Wain(string[] args)
...
+                var dllName = c.Option("-n |--name", "The name for the generated DLL (executable not supported at this time).  Defaults to the name of the script.", CommandOptionType.SingleValue);
+                var dllOption = c.Option("--dll", "Publish to a .dll instead of an executable.", CommandOptionType.NoValue);
+                var commandConfig = c.Option("-c | --configuration <configuration>", "Configuration to use for publishing the script [Release/Debug]. Default is \"Debug\"", CommandOptionType.SingleValue);
                 var publishDebugMode = c.Option(DebugFlagShort + " | " + DebugFlagLong, "Enables debug output.", CommandOptionType.NoValue);
-                var runtime = c.Option("-r |--runtime", "The runtime to publish the self contained EXE to. Defaults to the your current platform.", CommandOptionType.SingleValue);
+                var runtime = c.Option("-r |--runtime", "The runtime used when publishing the self contained executable. Defaults to the your current runtime.", CommandOptionType.SingleValue);
```
seesharper: Minor typo "Defaults to the your current runtime." should be "Defaults to your current runtime." 😎

seesharper:
Awesome 👍

PR #325, opened 2018-07-20
Feature/single netcoreapp target

seesharper:
This PR removes support for netcoreapp2.0 meaning that we only target netcoreapp2.1 for the dotnet script executable. 
Since OmniSharp is Net461 now, we can also stop cross compiling Dotnet.Script.DepenencyModel for both netstandard2.0 and net46.

This also means that users will get the same target version regardless of whether they've installed using chocolatey or through the global tools.


filipw:
LGTM

PR #328, opened 2018-08-16
Feature/better logging

seesharper:
Adds support for using `Microsoft.Extensions.Logging` for writing log messages to the console. 
As before log messages are written to `stderr` and hence the custom implementations of various interfaces from `Microsoft.Extensions.Logging` since there is no built-in support for writing to `stderr`.

We continue to only reference the actual logging library in the host project and that means that `Dotnet.Script.Core` and `Dotnet.Script.DepenencyModel` continue to just rely on its own logging abstraction. 

We have also added the `--verbosity` option that makes it possible to specify the logging level.

```
 --verbosity Set the verbosity level of the command. Allowed values are t[trace], d[ebug], i[nfo], w[arning], e[rror], and c[ritical].
```

In addition to these changes, all the tests have been refactored to use the `ScriptTestRunner` class and that removed a lot of duplication of code that gathered arguments and executed the `dotnet-script` process. 

When debugging fixtures in process, the log messages are now written to the test output window.


src/Dotnet.Script.DependencyModel/Context/ProfiledRestorer.cs
```diff
@@ -0,0 +1,26 @@
+﻿using System.Diagnostics;
+using Dotnet.Script.DependencyModel.Logging;
+
+namespace Dotnet.Script.DependencyModel.Context
+{
+    public class ProfiledRestorer : IRestorer
```
filipw: 👍

src/Dotnet.Script/LogHelper.cs
```diff
@@ -0,0 +1,128 @@
...
+        }
+
+        public void Flush()
+        {            
+            // No action required as for every write, data is sent directly to the console
+            // output stream
+        }
+    }
+
+    public class AnsiSystemErrorConsole : IAnsiSystemConsole
```
filipw: why would we skip colors for non-windows?

src/Dotnet.Script.DependencyModel/Logging/LogActionExtensions.cs
```diff
@@ -1,22 +1,69 @@
...
+        public static LogLevel FromString(string levelName)
+        {
+            if (string.IsNullOrWhiteSpace(levelName))
+            {
+                return LogLevel.Warning;
+            }
+
+            if (!_levelMap.TryGetValue(levelName, out var level))
+            {
+                throw new InvalidOperationException($"Unknown level name : {levelName}");
```
filipw: should this throw or maybe better to silently fallback to some default log level like Info. it would not be great for logging config to crash the app?

src/Dotnet.Script/LogHelper.cs
seesharper: This just just a copy of an internal class.
https://github.com/aspnet/Logging/blob/c9a97e95004f27abff8012bca1ece1f2258eab90/src/Microsoft.Extensions.Logging.Console/ConsoleLogger.cs#L289

I believe the actual coloring is handled in the `AnsiLogConsole` class
https://github.com/aspnet/Logging/blob/master/src/Microsoft.Extensions.Logging.Console/Internal/AnsiLogConsole.cs

Issue #306, opened 2018-05-31
Shebang not working with runtime directive in script file


Gorea235:
I'm trying to play around with using these scripts, and I'm attempting to allow the file to be run directly from the terminal, however whenever I use the shebang and runtime directive together, I get the following error message:

`test.csx(2,1): error CS1024: Preprocessor directive expected`

I'm using the following script:

```csharp
#! /usr/bin/env dotnet script
#! "netcoreapp2.1"

Console.WriteLine("Hello world");
```

If I remove the runtime directive, using `./test.csx` works fine, and if it remove the shebang instead, then `dotnet script test.csx` works fine, it's just when both are present, `dotnet script` seems to not like it. I've tried putting the shebang in with double quote around it, but then `dotnet script` just tries to load that as a runtime, and it doesn't seem to work out.

Gorea235:
I think I accidentally narrowed down the problem. If you remove the shebang, but leave the version directive on the second line, it does the same thing.

filipw:
This is really Roslyn behavior, it only allows a single of those at the moment, and only at the top of the file.

That said, you can likely get rid of `#! "netcoreapp2.1"`. It's used only to instruct OmniSharp that this script should get language services in .NET Core 2.1 format, but there is now a new feature in OmniSharp that allows you to do that by `omnisharp.json` instead (locally or globally) https://github.com/OmniSharp/omnisharp-roslyn/pull/1154

Therefore I think we may even remove this. At runtime it has no meaning because at runtime the runner dictates the TFM and in the case of a global tool it's `netcoreapp2.1`.

Gorea235:
Alright, thank you! It might be worth removing it in that case, or at least specifying this in the documentation. I will just use the scripts without the framework directive, so the issue shouldn't be a problem.

seesharper:
We should maybe consider scaffolding the shebang on *nix. We already set up the `omnisharp.json` during `dotnet script init` so OmniSharp will be happy either way

tomlm:
I can't seem to get shebang to work on *nix at all.  Anyone have an example which works?

tomlm:
#345 is new PR for this (I messed up the branch for #341.
2 things to make shebang work:

1. just add the single directive **!#/usr/bin/env dotnet-script
2. mark the file as executable with **chmod +x foo.csx**
3. make sure file doesn't have Byte Order Mark (BOM) and has Linux LF, not windows 

[...]

PR #341, opened 2018-09-25
make csx files directly runnable on Windows, OSX and Linux

tomlm:
**On windows**
* change dotnet-init to register .csx files to be handled by dotnet-script in the registry

**On Mac/Linux**
* change dotnet-init to add a donet-script shebang instruction to the csx files on init
* change dotnet-init mark the new csx file as executable via chmod command
* strip the shebang off the source code before sending to the compiler

### DIFF
--- a/src/Dotnet.Script.Core/Scaffolder.cs
+++ b/src/Dotnet.Script.Core/Scaffolder.cs
@@ -1,6 +1,7 @@
 ﻿using Dotnet.Script.Core.Templates;
 using Dotnet.Script.DependencyModel.Environment;
 using Newtonsoft.Json.Linq;
+using System;
 using System.Diagnostics;
 using System.IO;
 using System.Linq;
@@ -43,10 +44,12 @@ namespace Dotnet.Script.Core
                     RuntimeInformation.IsOSPlatform(OSPlatform.Linux))
                 {
                     // add a shebang to set dotnet-script as the interpreter for .csx files
-                    scriptFileTemplate = "#!/usr/bin/env dotnet-script\n" + scriptFileTemplate;
+                    // add make sure we are using environment newlines, because shebang won't work with windows cr\lf
+                    scriptFileTemplate = "#!/usr/bin/env dotnet-script" + Environment.NewLine +
+                        scriptFileTemplate.Replace("\r\n", Environment.NewLine);
                 }
 
-                File.WriteAllText(pathToScriptFile, scriptFileTemplate);
+                File.WriteAllText(pathToScriptFile, scriptFileTemplate, System.Text.Encoding.ASCII /* Linux shebang can't handle BOM */);
 
                 if (RuntimeInformation.IsOSPlatform(OSPlatform.OSX) ||
                     RuntimeInformation.IsOSPlatform(OSPlatform.Linux))
@@ -133,8 +136,20 @@ namespace Dotnet.Script.Core
             string dotnetScriptPath = Path.Combine(installLocation, "dotnet-script.dll").Replace(@"\", "/");
             if (!File.Exists(pathToLaunchFile))
             {
-                string lauchFileTemplate = TemplateLoader.ReadTemplate("launch.json.template");
-                string launchFileContent = lauchFileTemplate.Replace("PATH_TO_DOTNET-SCRIPT", dotnetScriptPath);
+                string launchFileTemplate = null;
+                if (RuntimeInformation.IsOSPlatform(OSPlatform.OSX) ||
+                    RuntimeInformation.IsOSPlatform(OSPlatform.Linux))
+                {
+                    // since we have the shebang in the .csx file, we have to massage it to remove it when compiling
+                    // if it is massaged the debugger won't debug it because it detects it's changed, so this launch
+                    // template sets the flag which says it's ok to debug anyway
+                    launchFileTemplate = TemplateLoader.ReadTemplate("launch-linux.json.template");
+                }
+                else if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
+                {
+                    launchFileTemplate = TemplateLoader.ReadTemplate("launch.json.template");
+                }
+                string launchFileContent = launchFileTemplate.Replace("PATH_TO_DOTNET-SCRIPT", dotnetScriptPath);
                 File.WriteAllText(pathToLaunchFile, launchFileContent);
                 _scriptConsole.WriteSuccess($"...'{pathToLaunchFile}' [Created]");
             }
--- a/src/Dotnet.Script.Core/Scaffolder.cs
+++ b/src/Dotnet.Script.Core/Scaffolder.cs
@@ -1,11 +1,14 @@
 ﻿using Dotnet.Script.Core.Templates;
 using Dotnet.Script.DependencyModel.Environment;
+using Dotnet.Script.DependencyModel.Logging;
+using Dotnet.Script.DependencyModel.Process;
 using Newtonsoft.Json.Linq;
 using System;
 using System.Diagnostics;
 using System.IO;
 using System.Linq;
 using System.Runtime.InteropServices;
+using System.Text;
 using System.Text.RegularExpressions;
 
 namespace Dotnet.Script.Core
@@ -15,9 +18,11 @@ namespace Dotnet.Script.Core
         private ScriptEnvironment _scriptEnvironment;
         private const string DefaultScriptFileName = "main.csx";
         private ScriptConsole _scriptConsole = ScriptConsole.Default;
+        private CommandRunner _commandRunner;
 
-        public Scaffolder()
+        public Scaffolder(LogFactory logFactory)
         {
+            _commandRunner = new CommandRunner(logFactory);
             _scriptEnvironment = ScriptEnvironment.Default;
         }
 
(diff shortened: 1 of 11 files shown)


src/Dotnet.Script/Program.cs
```diff
@@ -267,21 +282,19 @@ private static async Task<int> RunScript(string file, bool debugMode, LogFactory
             var absoluteFilePath = Path.IsPathRooted(file) ? file : Path.Combine(Directory.GetCurrentDirectory(), file);
             var directory = Path.GetDirectoryName(absoluteFilePath);
 
-            using (var filestream = new FileStream(absoluteFilePath, FileMode.Open, FileAccess.Read, FileShare.ReadWrite))
+            var rawCode = File.ReadAllText(absoluteFilePath);
+            var sourceText = SourceText.From(RemoveShebang(rawCode));
```
seesharper: Could this cause some issues with breakpoints since we alter the original "source" ?

src/Dotnet.Script.Core/Scaffolder.cs
```diff
@@ -100,12 +113,26 @@ private void CreateLaunchConfiguration(string currentWorkingDirectory)
                 Directory.CreateDirectory(vsCodeDirectory);
             }
 
+            // on windows we use this as opportunity to make the file associate for .csx -> dotnet-script
+            // (If dotnet install gives us an install time hook this code should move there)
+            if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
```
seesharper: Same here, could use `_scriptEnvironment.IsWindows`

src/Dotnet.Script.Core/Scaffolder.cs
```diff
@@ -100,12 +113,26 @@ private void CreateLaunchConfiguration(string currentWorkingDirectory)
...
+            {
+                // register dotnet-script as the tool to process .csx files
+                string cmdPath = Path.Combine(Path.GetTempPath(), "register.cmd");
+                using (var textWriter = new StreamWriter(cmdPath))
+                {
+                    textWriter.WriteLine("@reg add HKCU\\Software\\classes\\.csx /f /ve /t REG_SZ -d dotnetscript > nul");
+                    textWriter.WriteLine("@reg add HKCU\\Software\\Classes\\dotnetscript\\Shell\\Open\\Command /f /ve /t REG_EXPAND_SZ /d \"%%USERPROFILE%%\\.dotnet\\tools\\dotnet-script.exe %%1 -- %%*\" > nul");
```
seesharper: dotnet-script can also be installed using Chocolatey or install.ps1/install.sh and then the install location is different.
`ScriptEnvironment` provides the install location. Maybe use that?

src/Dotnet.Script/Program.cs
tomlm: hmmm...great question, I will explore

seesharper: What happens if we just leave it as is? We used to have `#! "netcoreapp2.1"` and the compiler does not care about that?

filipw:
> * strip the shebang off the source code before sending to the compiler

this shouldn't be needed - Roslyn supports shebang in the source code https://github.com/dotnet/roslyn/blob/fab7134296816fc80019c60b0f5bef7400cf23ea/src/Compilers/CSharp/Portable/Syntax/DirectiveTriviaSyntax.cs#L53

src/Dotnet.Script.Core/Scaffolder.cs
```diff
@@ -1,46 +1,59 @@
...
+                    scriptFileTemplate = "#!/usr/bin/env dotnet-script\n" + scriptFileTemplate;
+                }
+
                 File.WriteAllText(pathToScriptFile, scriptFileTemplate);
+
+                if (RuntimeInformation.IsOSPlatform(OSPlatform.OSX) ||
+                    RuntimeInformation.IsOSPlatform(OSPlatform.Linux))
+                {
+                    // mark .csx file as executable, this activates the shebang to run dotnet-script as interpreter
+                    Process.Start($"/bin/chmod", $"+x {pathToScriptFile}").WaitForExit();
```
filipw: I think it would make sense to spit out there errors from these external processes, like https://github.com/filipw/dotnet-script/blob/de824d38292f3ca00563bb5e9299de888b4be38f/src/Dotnet.Script.DependencyModel/Process/CommandRunner.cs#L54

src/Dotnet.Script/Program.cs
tomlm: Here's the funky thing, if you have JUST the shebang the compiler don't care and it works fine.
 
If you have shebang AND the #! "netcoreapp2.1" then the compiler barfs on the shebang. 
I just updated it so that the shebang is replaced and extra CR is added so line numbers remain the same.

src/Dotnet.Script.Core/Scaffolder.cs
tomlm: oh, snap, k I'll do that

src/Dotnet.Script.Core/Scaffolder.cs
tomlm: can I just use the command runner?

tomlm:
They lie about supporting it, see previous comment.
If you have only shebang it's fine
If you both have shebang AND #! "netcoreapp1."  directive then it blows up.

I would love to drop the #! directive from the template and just have the shebang, then I could drop all of the shenanigans. Is that #! "netcoreapp2" directive important?  It's really unfortunately it has the same name signature as the shebang

src/Dotnet.Script.Core/Scaffolder.cs
seesharper: Yup, that is okay. If fact, better :)

seesharper:
Yes, that is correct. The compiler accepts only one `#!` directive.  The reason for `#! netcoreapp2.1` is to inform `OmniSharp` that it should resolve compilation dependencies in the context of an `netcoreapp2.1`.  We actually don't need this hint anymore as we scaffold this into the `omnisharp.json` file during `dotnet script init`. @filipw A good time to get rid of scaffolding `#! "netcoreapp2.1"` in the scripts?

tomlm:
the more I think about it the more I feel like this is a mac/Linux only thing and users can pick to have the shebang or not, they just get one thing.  It super cleans up the code and I don't think it's even necessarily wrong.

filipw:
> Yes, that is correct. The compiler accepts only one `#!` directive. The reason for `#! netcoreapp2.1` is to inform `OmniSharp` that it should resolve compilation dependencies in the context of an `netcoreapp2.1`. We actually don't need this hint anymore as we scaffold this into the `omnisharp.json` file during `dotnet script init`. @filipw A good time to get rid of scaffolding `#! "netcoreapp2.1"` in the scripts?

yeah I think it should go away - see https://github.com/filipw/dotnet-script/issues/306#issuecomment-393631656

src/Dotnet.Script/Program.cs
seesharper: Guess we don't have to touch the source if we stick we just one shebang directive 👍

seesharper:
> remove shebangaigans

So that is the plural form. Always wondered about that 😂

tomlm:
argh, I was trying to merge 2 branches locally and I messed up this branch ! :( :( :(

src/Dotnet.Script.Core/Scaffolder.cs
```diff
@@ -100,14 +115,28 @@ private void CreateLaunchConfiguration(string currentWorkingDirectory)
...
+            if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
+            {
+                // register dotnet-script as the tool to process .csx files
+                string cmdPath = Path.Combine(Path.GetTempPath(), "register.cmd");
+                using (var textWriter = new StreamWriter(cmdPath))
+                {
+                    textWriter.WriteLine("@reg add HKCU\\Software\\classes\\.csx /f /ve /t REG_SZ -d dotnetscript > nul");
+                    textWriter.WriteLine($"@reg add HKCU\\Software\\Classes\\dotnetscript\\Shell\\Open\\Command /f /ve /t REG_EXPAND_SZ /d \"{_scriptEnvironment.InstallLocation}\\dotnet-script.cmd %%1 -- %%*\" > nul");
+                }
+                Process.Start("cmd.exe", $"/c {cmdPath}").WaitForExit();
```
seesharper: I would use the `CommandRunner` here as we will get debug logging for free

src/Dotnet.Script.Core/Scaffolder.cs
```diff
@@ -1,46 +1,61 @@
...
+
+                if (RuntimeInformation.IsOSPlatform(OSPlatform.OSX) ||
+                    RuntimeInformation.IsOSPlatform(OSPlatform.Linux))
+                {
+                    // add a shebang to set dotnet-script as the interpreter for .csx files
+                    // and make sure we are using environment newlines, because shebang won't work with windows cr\lf
+                    scriptFileTemplate = $"#!/usr/bin/env dotnet-script" + Environment.NewLine + scriptFileTemplate.Replace("\r\n", Environment.NewLine);
+                }
+
+                File.WriteAllText(pathToScriptFile, scriptFileTemplate, System.Text.Encoding.ASCII /* Linux shebang can't handle BOM */);
```
seesharper: It is possible to write using UTF8 without the BOM 👍

```
Encoding utf8WithoutBom = new UTF8Encoding(false);
```

tomlm:
This branch is messed up. I am creating new branch and PR for it.

PR #342, opened 2018-09-25
Speed up scripts by auto-publishing to DLL cache

tomlm:
This is identical to previous PR, except I separated out the extension registry stuff into separate PR #341 as requested.

### DIFF
--- a/src/Dotnet.Script/Program.cs
+++ b/src/Dotnet.Script/Program.cs
@@ -1,3 +1,4 @@
+#define AUTOPUBLISH
 using System;
 using System.Collections.Generic;
 using System.Diagnostics;
@@ -5,6 +6,7 @@ using System.IO;
 using System.Linq;
 using System.Reflection;
 using System.Runtime.InteropServices;
+using System.Security.Cryptography;
 using System.Text;
 using System.Threading.Tasks;
 using Dotnet.Script.Core;
@@ -74,7 +76,7 @@ namespace Dotnet.Script
 
             var verbosity = app.Option("--verbosity", " Set the verbosity level of the command. Allowed values are t[trace], d[ebug], i[nfo], w[arning], e[rror], and c[ritical].", CommandOptionType.SingleValue);
 
-            CommandOption register=null;
+            CommandOption register = null;
             if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
             {
                 register = app.Option("--register", "Register .csx file extension to be executable by dotnet-script.", CommandOptionType.NoValue);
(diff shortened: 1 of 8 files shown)


tomlm:
NOTE: On this change https://github.com/filipw/dotnet-script/pull/342/commits/639c51855cde39ed9e973a7bdf728e75c6922009
hmm.. I had to do this to make the ScriptPackagesTests pass, but I don't think this behavior was my changes but instead I merged latest master with my delta.  The paths being generated had a mixture of forward and backslashes causing the code which was looking for the assemblies to fail.  This fixes it in the locale, but I couldn't figure out where these paths were being created, and I think a better fix is to find out who is generating mixed slash paths.
Anyone understand this part of the code?.

src/Dotnet.Script/Program.cs
```diff
@@ -232,20 +238,104 @@ private static int Wain(string[] args)
...
+                        {
+                            absoluteSourcePath = Path.IsPathRooted(file.Value) ? Path.GetFullPath(file.Value) : Path.GetFullPath(Path.Combine(Directory.GetCurrentDirectory(), file.Value));
+                            code = SourceText.From(File.ReadAllText(absoluteSourcePath));
+                        }
+
+                        // given the path to a script we create a %temp%\dotnet-scripts\{uniqueFolderName} path
+                        string pathToDll = Path.Combine(publishDirectory, Path.GetFileNameWithoutExtension(absoluteSourcePath) + ".dll");
+
+                        // source hash is the checkSum of the code
+                        string sourceHash = Convert.ToBase64String(code.GetChecksum().ToArray());
```
seesharper: The only thing that I am worried about here is NuGet packages referenced without version numbers. 
```
#r "nuget: SomePackage"
```
At some point `SomePackage` gets bumped while we still run with the old version.

src/Dotnet.Script/Program.cs
```diff
@@ -232,20 +238,104 @@ private static int Wain(string[] args)
...
+                        var optimizationLevel = OptimizationLevel.Debug;
+                        if (configuration.HasValue() && configuration.Value().ToLower() == "release")
+                        {
+                            optimizationLevel = OptimizationLevel.Release;
+                        }
+                        exitCode = await RunScript(file.Value, debugMode.HasValue(), logFactory, optimizationLevel, app.RemainingArguments.Concat(argsAfterDoubleHypen), interactive.HasValue(), packageSources.Values?.ToArray());
+                    }
+                    else
+                    {
+                        string cacheFolder = Path.Combine(Path.GetTempPath(), "dotnet-scripts");
```
seesharper: At some point we need to move all this logic out of `Program.cs` so that it can be used for people using `dotnet-script` programatically. This applies to everything here so we can do that later 👍

src/Dotnet.Script/Program.cs
filipw: I think this is an acceptable tradeoff. 
I also think at some point in the future we will introduce some global JSON settings file so you could preset certain flags (i.e. --no-cache) globally so that should help

src/Dotnet.Script/Program.cs
filipw: I agree - let's keep this PR limited in size and put that in the next steps

src/Dotnet.Script/Program.cs
seesharper: Should we at the very least offer a way to clear the cache? 

`dotnet script clear-cache` 

I don't know if that should clear all cached scripts or just the "current" one ?

seesharper: @filipw thoughts?

tomlm: it also seems like something that more clever caching of the project file would allow us to decide that it is time to recompile.

tomlm: I have updated the README to describe the dll caching behavior

tomlm:
I don't understand why the appveyor is failing.  All of the tests pass on my machine, and it appears to be just one test which never completes on the server.  Really weird.  I don't know how to debug this.

seesharper:
Build restarted. Let's see how that turns out

seesharper:
FYI. Same issue locally on my windows VM

tomlm:
woot! Thanks guys!

PR #345, opened 2018-09-26
make csx files directly runnable on Windows, OSX and Linux

tomlm:
* register .csx extensions on windows
* change scaffolding to add shebang and mark as executable on osx/Linux
* update readme to describe way of invoking scripts

### DIFF
--- a/src/Dotnet.Script.Core/Scaffolder.cs
+++ b/src/Dotnet.Script.Core/Scaffolder.cs
@@ -1,38 +1,41 @@
-﻿using System;
-using System.IO;
-using System.Linq;
-using System.Reflection;
-using System.Text.RegularExpressions;
-using Dotnet.Script.Core.Templates;
+﻿using Dotnet.Script.Core.Templates;
 using Dotnet.Script.DependencyModel.Environment;
 using Dotnet.Script.DependencyModel.Logging;
-using Dotnet.Script.DependencyModel.ProjectSystem;
+using Dotnet.Script.DependencyModel.Process;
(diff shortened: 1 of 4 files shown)


tomlm:
NOTE: I had messed up my old branch so bad I just created this new branch to apply the changes cleanly

src/Dotnet.Script.Core/Scaffolder.cs
```diff
@@ -100,14 +119,27 @@ private void CreateLaunchConfiguration(string currentWorkingDirectory)
                 Directory.CreateDirectory(vsCodeDirectory);
             }
 
+            // on windows we use this as opportunity to make the file association for .csx -> dotnet-script
+            // (If/when dotnet install command provides us an install time hook this code should move there)
+            if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
+            {
+                // check to see if .csx is mapped to dotnet-script
+                if (_commandRunner.Execute("reg", "query HKCU\\Software\\classes\\.csx") != 0)
```
seesharper: I assume we always have the necessary permission to write to the registry?

tomlm: yeah it works under user permissions

tomlm:
I added a test that we can execute the script directly, which will fail if we don't have the whole shebang. (lol)

seesharper:
@tomlm Looks like it fails on *nix 😸

tomlm:
So when do you guys do releases?

PR #349, opened 2018-10-01
Add static modifier to Program

atifaziz:
Because it doesn't have an instance-based state or methods and marking `Program` static makes that very clear and should be kept so.


filipw:
thanks!

PR #350, opened 2018-10-01
Make case-insensitive string checks deterministic

atifaziz:
Because [`String.ToLower`](https://docs.microsoft.com/en-us/dotnet/api/system.string.tolower?view=netstandard-2.0#System_String_ToLower) is, per its documentation, culture-specific:

> ## Remarks
>
> This method takes into account the casing rules of the current culture.

### DIFF
--- a/src/Dotnet.Script/Program.cs
+++ b/src/Dotnet.Script/Program.cs
@@ -69,6 +69,7 @@ namespace Dotnet.Script
             var interactive = app.Option("-i | --interactive", "Execute a script and drop into the interactive mode afterwards.", CommandOptionType.NoValue);
 
             var configuration = app.Option("-c | --configuration <configuration>", "Configuration to use for running the script [Release/Debug] Default is \"Debug\"", CommandOptionType.SingleValue);
+            var release = StringComparer.OrdinalIgnoreCase.PredicateEquals("release");
 
             var packageSources = app.Option("-s | --sources <SOURCE>", "Specifies a NuGet package source to use when resolving NuGet packages.", CommandOptionType.MultipleValue);
 
@@ -98,7 +99,7 @@ namespace Dotnet.Script
                     if (!string.IsNullOrWhiteSpace(code.Value))
                     {
                         var optimizationLevel = OptimizationLevel.Debug;
-                        if (configuration.ValueEquals("release", StringComparison.OrdinalIgnoreCase))
+                        if (configuration.ValueEquals(release))
                         {
                             optimizationLevel = OptimizationLevel.Release;
                         }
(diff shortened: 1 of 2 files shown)


src/Dotnet.Script/Program.cs
```diff
@@ -98,7 +98,7 @@ private static int Wain(string[] args)
                     if (!string.IsNullOrWhiteSpace(code.Value))
                     {
                         var optimizationLevel = OptimizationLevel.Debug;
-                        if (configuration.HasValue() && configuration.Value().ToLower() == "release")
+                        if (configuration.ValueEquals("release", StringComparison.OrdinalIgnoreCase))
```
seesharper: `StringComparison.OrdinalIgnoreCase` could be the default here. That would clean up the call site ever more

atifaziz: That can be done post-PR as you see fit? Meanwhile, I didn't want to burn a global policy into an extension method.

seesharper: I was thinking setting it as a default value for the parameter. That way it is not "burned" into the method :)

atifaziz: Took a slightly different approach in 8fd017d3cb2a45aae5bfc904f7ae970cf736a25c, which kills two birds with one stone (the second bird is the `"release"` literal).

atifaziz:
`StringComparison.OrdinalIgnoreCase` is what I had [in the first place](https://github.com/filipw/dotnet-script/pull/350/commits/3521b2c080fb5a67882833b61adc7e30c1ae5f6b). I can revert if you find it simpler.

> I am also not sure this fixes any bug, which even further makes me wonder why would we need it.

It's never a bug until it is one. 😄 This is about good practice and erring on the safe side, so yeah, you avoid the bug in advance because we have the methods and it's easy to do so.

> I mean does actually `<ASCII INPUT FROM CONSOLE>.ToLower() == "release"` break anywhere?

Not specifically with _release_ perhaps (though I don't know case comparison rules of all cultures), but the classic test/example is not to break on a Turkish box. For instance, the following returns `false`:

> StringComparer.Create(new CultureInfo("tr-TR"), true).Equals("FILE", "file")

Both, `"FILE"` and `"file"` are _ASCII_, but Turkish has [four versions of the letter I](https://en.wikipedia.org/wiki/Dotted_and_dotless_I) that don't translate to upper- and lowercase the same way as in English. Rather than worry, why not just err on the safe side?

> a lot less maintainable now

Removing four duplicate comparisons to a configuration named _release_ is less maintainable than no repeating code? But if `+28 −4` is the metric/measure of maintenance then I've pooped. 💩 

> I'd rather see this fixed by calling `ToLowerInvariant()` or `StringComparison.OrdinalIgnoreCase`, just for the sake of simplicity.

I can settle for this if the functional approach feels too weird. :wink:

filipw:
great, then let's revert to the original approach

filipw:
thanks

PR #353, opened 2018-10-01
Use just the content hash for cached publication

atifaziz:
This builds on top of the great addition in PR #342 by @tomlm.

I didn't understand the benefits of maintaining two hashes in #342, one of the script's path and the other of the script's content. This PR uses just the latter (hash of the script content) as part of the path to the cached DLL, yielding the following advantages:

1. The script can be moved or copied around the file system and the cache will be hit so long as it's the same content!

2. It avoids locking issues when updating the source of long-running scripts (which includes simple scripts having a call to `Console.ReadLine()`at the end). For example, if a script takes some time to run, I'm not able to edit and update it. The implementation in #345 fails due to locking issues, at least on Windows. With this PR, the change in the content will mean a separately published/cached version will run without interfering with existing one!

These advantages are even more pronounced when you think of a repo of C# scripts where you can switch between branches & run different versions of scripts without affecting those already in flight. The cache will help even more when switching between different versions.

This PR also changes the encoding of the hash bytes in text from Base64 to the more conventional hexadecimal. If it needs to be made shorter then [Base32](https://crockford.com/wrmg/base32.html) would be a better compromise.

### DIFF
--- a/src/Dotnet.Script/Extensions.cs
+++ b/src/Dotnet.Script/Extensions.cs
@@ -4,9 +4,5 @@ namespace Dotnet.Script
 {
     static class Extensions
     {
-        public static string ToHexadecimalString(this byte[] bytes) =>
-            BitConverter.ToString(bytes)
-                        .Replace("-", string.Empty)
-                        .ToLowerInvariant();
     }
 }
--- a/src/Dotnet.Script/Extensions.cs
+++ b/src/Dotnet.Script/Extensions.cs
@@ -1,0 +1,12 @@
+using System;
+
+namespace Dotnet.Script
+{
+    static class Extensions
+    {
+        public static string ToHexadecimalString(this byte[] bytes) =>
+            BitConverter.ToString(bytes)
+                        .Replace("-", string.Empty)
+                        .ToLowerInvariant();
+    }
+}
--- a/src/Dotnet.Script/Extensions.cs
+++ b/src/Dotnet.Script/Extensions.cs
@@ -1,8 +1,18 @@
-using System;
+using System.IO;
 
 namespace Dotnet.Script
 {
+    using System;
+
     static class Extensions
     {
+        public static bool CreateIfMissing(this DirectoryInfo directory)
+        {
+            if (directory == null) throw new ArgumentNullException(nameof(directory));
+            if (directory.Exists)
+                return false;
+            directory.Create();
+            return true;
+        }
     }
 }
--- a/src/Dotnet.Script/Extensions.cs
+++ b/src/Dotnet.Script/Extensions.cs
@@ -1,9 +1,8 @@
+using System;
 using System.IO;
 
 namespace Dotnet.Script
 {
-    using System;
-
     static class Extensions
     {
         public static bool CreateIfMissing(this DirectoryInfo directory)
--- a/src/Dotnet.Script/Program.cs
+++ b/src/Dotnet.Script/Program.cs
@@ -251,31 +251,15 @@ namespace Dotnet.Script
                     }
                     else
                     {
-                        string cacheFolder = Path.Combine(Path.GetTempPath(), "dotnet-scripts");
-                        // create unique folder name based on the path
-                        string uniqueFolderName = "";
-                        using (var sha = SHA256.Create())
-                        {
-                            uniqueFolderName = sha.ComputeHash(Encoding.UTF8.GetBytes(file.Value))
-                                                  .ToHexadecimalString();
-                        }
-
-                        string publishDirectory = Path.Combine(cacheFolder, uniqueFolderName);
-                        if (!Directory.Exists(publishDirectory))
-                        {
-                            Directory.CreateDirectory(publishDirectory);
-                        }
-
-                        string absoluteSourcePath;
+                        string absoluteSourcePath = null;
+                        Uri scriptUrl = null;
                         SourceText code;
                         if (!File.Exists(file.Value))
                         {
-                            if (IsHttpUri(file.Value))
+                            if (TryParseHttpUrl(file.Value, out scriptUrl))
                             {
                                 var downloader = new ScriptDownloader();
                                 var rawCode = await downloader.Download(file.Value);
-                                absoluteSourcePath = Path.Combine(publishDirectory, "source.csx");
-                                File.WriteAllText(absoluteSourcePath, rawCode);
                                 code = SourceText.From(rawCode);
                             }
                             else
@@ -289,23 +273,45 @@ namespace Dotnet.Script
                             code = SourceText.From(File.ReadAllText(absoluteSourcePath));
                         }
 
-                        // given the path to a script we create a %temp%\dotnet-scripts\{uniqueFolderName} path
-                        string pathToDll = Path.Combine(publishDirectory, Path.GetFileNameWithoutExtension(absoluteSourcePath) + ".dll");
-
                         // source hash is the checkSum of the code
-                        string sourceHash = code.GetChecksum().ToArray().ToHexadecimalString();
+                        string sourceHash =
+                            BitConverter.ToString(code.GetChecksum().ToArray())
+                                        .Replace("-", string.Empty)
+                                        .ToLowerInvariant();
+
+                        string cacheFolder = Path.Combine(Path.GetTempPath(), "dotnet-scripts");
+                        string publishDirectory = Path.Combine(cacheFolder, sourceHash);
+
+                        if (scriptUrl != null) // was downloaded?
+                        {
+                            var filename = scriptUrl.AbsolutePath.Split('/').Last();
+                            if (!filename.EndsWith(".csx", StringComparison.OrdinalIgnoreCase)
+                                || filename.IndexOfAny(Path.GetInvalidFileNameChars()) >= 0)
+                            {
+                                filename = "script.csx";
+                            }
+                            absoluteSourcePath = Path.Combine(publishDirectory, filename);
+                            var encoding = new UTF8Encoding(encoderShouldEmitUTF8Identifier: false);
+                            using (var writer = new StreamWriter(absoluteSourcePath, append: false, encoding))
+                            {
+                                code.Write(writer);
+                            }
+                        }
+
+                        // given the path to a script we create a %temp%\dotnet-scripts\{uniqueFolderName} path
+                        string pathToDll = Path.Combine(publishDirectory, "script.dll");
 
                         // get hash code from previous run 
-                        string hashCache = Path.Combine(publishDirectory, ".hash");
                         var compiler = GetScriptCompiler(true, logFactory);
 
-                        // if we don't have hash
-                        if (!File.Exists(hashCache) ||
-                             // or we haven't created a dll
-                             !Directory.Exists(publishDirectory) ||
-                             // the hashcode has changed (meaning new content)
-                             File.ReadAllText(hashCache) != sourceHash)
+                        // if we haven't created a dll
+                        if (!File.Exists(pathToDll))
                         {
+                            if (!Directory.Exists(publishDirectory))
+                            {
+                                Directory.CreateDirectory(publishDirectory);
+                            }
+
                             // then we autopublish into the %temp%\dotnet-scripts\{uniqueFolderName} path
                             var optimizationLevel = OptimizationLevel.Debug;
                             if (configuration.HasValue() && configuration.Value().ToLower() == "release")
@@ -320,9 +326,6 @@ namespace Dotnet.Script
 
                             // create the assembly in our cache folder
                             publisher.CreateAssembly<int, CommandLineScriptGlobals>(context, logFactory, Path.GetFileNameWithoutExtension(pathToDll));
-
-                            // save sourceHash for next time, so we can know it's ok to use the generated dll next time
-                            File.WriteAllText(hashCache, sourceHash);
                         }
 
 
@@ -347,7 +350,7 @@ namespace Dotnet.Script
         {
             if (!File.Exists(file))
             {
-                if (IsHttpUri(file))
+                if (TryParseHttpUrl(file, out _))
                 {
                     var downloader = new ScriptDownloader();
                     var code = await downloader.Download(file);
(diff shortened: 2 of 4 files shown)


src/Dotnet.Script/Extensions.cs
```diff
@@ -0,0 +1,18 @@
+using System.IO;
+
+namespace Dotnet.Script
+{
+    using System;
```
filipw: can we move the usings outside to be consistent?

src/Dotnet.Script/Program.cs
```diff
@@ -288,23 +273,45 @@ private static int Wain(string[] args)
...
-                        string pathToDll = Path.Combine(publishDirectory, Path.GetFileNameWithoutExtension(absoluteSourcePath) + ".dll");
-
                         // source hash is the checkSum of the code
-                        string sourceHash = Convert.ToBase64String(code.GetChecksum().ToArray());
+                        string sourceHash =
+                            BitConverter.ToString(code.GetChecksum().ToArray())
+                                        .Replace("-", string.Empty)
+                                        .ToLowerInvariant();
+
+                        string cacheFolder = Path.Combine(Path.GetTempPath(), "dotnet-scripts");
```
filipw: let's use the folder name `dotnet-script`, as in the singular "product name"

atifaziz: I think @tomlm meant it as plural but if we want to singular and product-based then how about `dotnet-script/cache`? It would enable other things might end up in the temporary directory to side under the same umbrella.

src/Dotnet.Script/Extensions.cs
atifaziz: Done in 3316cc217ea8580c1014bb360f1c0ea776bcc94d.

atifaziz:
>  I keep thinking about what could go wrong here, but haven't figured out anything obvious

One thing that does come to mind is if a script hash- or pound-loads (`#load`) other scripts and one of those sub-scripts change but not the root then the cache doesn't get invalidated. I didn't check deeper in the code stack if that's somehow indirectly covered (e.g. by the source checksum) but I somehow have my doubts. For such cases, one would want a hash of the full tree of sources. That said, this PR is not worse off than PR #342 and my assumption was that if it was merged then these concerns were probably addressed there. If not, happy to consider them now, gather everyone's thoughts and get it right.

seesharper:
I briefly mentioned this before, but we might end up with a stale dependency graph is we have NuGet references without version numbers. If a new package version gets released, the cache will not be invalidated and we end up running with the previous version of the NuGet package.

seesharper:
@atifaziz I tinkered with this a while back and I think my plan was to add a `Cacheable` property to the `ProjectFile` class. That would of course require parsing the script files as you see in the  `ScriptProjectProvider`, but in my opinion it is worth it. We should also emit a log message when we are unable to cache.

src/Dotnet.Script/Program.cs
```diff
@@ -288,23 +273,45 @@ private static int Wain(string[] args)
...
+                        DirectoryInfo publishDirectory = new DirectoryInfo(Path.Combine(cacheFolder, sourceHash));
```
tomlm: >DirectoryInfo publishDirectory = new DirectoryInfo(Path.Combine(cacheFolder, sourceHash)); [](start = 24, length = 90)

GetCheckSum() appears to be MD5 hash of the contents.  While unlikely it is possible for that to have a collision, which would mean that you are running a different script accidently!  ! ! 

Normally I wouldn't be concerned, except if I run a script that I think is going to print out happy emoticons and it accidently runs a DLL which deletes the root of the HD I would be understandably upset!

Using the source path to get the temp folder + the MD5 hash of the contents greatly reduces this collision, because you would have to have a SHA256 collision on the path + the MD5 collision on the content for you to accidently run the wrong code.

atifaziz: @tomlm Thanks for dropping in and giving the background to why you chose two hashes (the path and the content). I had another theory but good to hear it straight from the horse's mouth (don't mean to call you a horse so I apologise in advance for my poor choice of this phrase).

> GetCheckSum() appears to be MD5 hash of the contents.

Where do you see that documented or how do you know? From what I can see in the Roslyn source code, [`SourceHashAlgorithm`](https://github.com/dotnet/roslyn/blob/Visual-Studio-2017-Version-15.8/src/Compilers/Core/Portable/Text/SourceHashAlgorithm.cs) doesn't event support MD5 and [SHA-1 is the default](https://github.com/dotnet/roslyn/blob/958f2354c4d83dbb0e7723d0a8079a0dfbc33f25/src/Compilers/Core/Portable/Text/SourceText.cs#L166-L171):

```c#
        public static SourceText From(
            Stream stream,
            Encoding encoding = null,
            SourceHashAlgorithm checksumAlgorithm = SourceHashAlgorithm.Sha1,
            bool throwIfBinaryDetected = false,
            bool canBeEmbedded = false)
```

If it's any consolation, Git uses SHA-1 for identity:

> Revision control systems such as Git, Mercurial, and Monotone use SHA-1 not for security but to identify revisions and to ensure that the data has not changed due to accidental corruption
>
> &mdash;<cite>[Data integrity, SHA-1, Wikipedia](https://en.wikipedia.org/wiki/SHA-1#Data_integrity)</cite>

If there's concern then we can switch the algorithm to [SHA-256](https://github.com/dotnet/roslyn/blob/Visual-Studio-2017-Version-15.8/src/Compilers/Core/Portable/Text/SourceHashAlgorithm.cs#L23).

As an additional measure, we can append or prepend the byte length of the source text to the source text itself before computing the hash or we can tag it on the hash. However, before we get too fancy, I think the real elephant (security-wise) is allowing easy execution of remote scripts than this:

> Normally I wouldn't be concerned, except if I run a script that I think is going to print out happy emoticons and it accidently runs the dll of a script which deletes the root of the HD I would be understandably upset!

I think this sort of problem is much more likely to happen through implementation/design oversight (which we are working through here), such as just hashing the root script or loading a NuGet package without a version number, than the hash collision. If one NuGet package version is print smileys and the next version deletes from the root of the HDD then I'd be equally upset. :wink:

We are now thinking of bypassing caching for scripts that don't pin their NuGet package references to specific versions.

atifaziz:
@seesharper I had to make `ScriptFilesResolver` deterministic for the purpose of caching. It was internally using a hash-set to build a unique list but the trouble with a hash-set is that is doesn't guarantee order so I had to maintain an ordered list as well.

I'm hoping to use `ScriptFilesResolver` to build the list of `#load`-ed files to be hashed together.

src/Dotnet.Script/Program.cs
tomlm: I don't think the algo is necessarily the issue, as much as doing things that are simply to descrease the chance for these things to happen. 

Unique inputs are
* **path**
* **contents** of file

I think the fundamental issue is any hashcode() can have a collision, and we need to look at the consequences of it.

to me, 

* **path** ==> identity of program, 
* **contents** => Has the program changed. 

I think my bias would be:
* cache folder name should be the **path** to the source file encoded into folder name. So c:\foo\boar\blat.csx ==> %tmp%\c_foo_boar_blat_csx\script.dll.  This removes hashing as the mechanism to identify the compiled code, as it's a strict algo to get to the folder name.

* Use **code.GetHashCode()** to determine if the code has changed.  The side-effect of getting a collision is executing old code, not executing someone else's program.

this feels pretty simple and something I would feel comfortable with.  


Thoughts?

atifaziz: > **path** ==> identity of program,
 
A path is generally not the identity of a file, even for file systems (think symbolic links) or version control systems like Git that call it an _object database_ (where the identity of each blob is its SHA-1 hash).

That's thinking at the file level but we need to think at the level of what constitutes a _program_ or _script_ as a whole; which comes down to 2 things:

1. The content of the file
2. The contents of `#load`-ed files

A script is the sum of the above and what should serve as the hash, identity or cache key. What's more, we decide that for a script to be eligible for caching, we need the NuGet references to be pinned to a specific version. This is to ensure that caching is avoided _unless_ the state of the script can be capture _deterministically_.

> The side-effect of getting a collision is executing old code, not executing someone else's program.

I think we need to go the distance here otherwise this will only get reported back as a bug. It'll need to be solved eventually so why not address it from the start? To me, a cache that's not invalidated reliably, especially with respect to _executable code_, is useless.

filipw: I think it can all be made a lot more simple if we drop this caching in the temp folder and just adopt the approach that FAKE and a number of other tools use, and that is to cache in a hidden directory, say `.dotnet-script`, relative to the executed script. The added benefit is that if you want, you can manually delete the cache by deleting the folder. Another benefit is, if you want to, you can easily distribute the script or commit to source control *with* the cached version alongside it. This would also allow you to move things around on your filesystem without affecting the cache.

And then the hash should be computed by looking at all syntax trees.

seesharper: It might be simpler to just create a hidden folder alongside with the script, but there are a couple of considerations that might be worth taking into account.

* Some IDE's, for instance visual studio code shows all folders including hidden folders and files. I really think we should strive for a solution where the end users only sees the script file and not the folder structure we create behind the scenes to make all this work.  It might be harder to do, that is one thing, but if the user moves a script from one location to another, a cache miss is probably not a bug issue. Executing from the wrong cache on the other hand is something we must avoid by all means. 
Personally, I would like to open my script folder and see nothing but the script in there. If I see hidden folders and files in there, it kind of takes the script out of scripting if you know what I mean :) 

* Source control

If the user want to add a script to git, we would have to provide the proper `.gitignore` entry to prevent the cache from being submitted to git.

atifaziz: @seesharper I second all your points. 💯

filipw: well, I do not agree with any of the points 🤣 that said I don't have a strong preference either - but I don't think that hiding DLLs from people changes much

atifaziz: I feel we are getting a little distracted here 😄 by putting into question several aspects of caching, like merits/choice of hashing algorithms, cache location and more. I go with the assumption that much of this was reviewed and debated since PR #342 _was merged_. However, I saw two flaws the implementation:

1. Inability to update long-running scripts due to locking of auto-published version
2. Incomplete cache invalidation (ignores changes in `#load`-ed files)

This PR is mostly about the first issue. I was thinking of addressing the second problem in another PR but happy to fold it here too if that's the right call by everyone.

For this PR, using the script path as part of the cache key seemed redundant and using the content solved a whole set of problems, avoided others _and_ simplified things (how often do you get those many wins? 😏). There seems to be some insecurity left about the collision of SHA-1 hashes that I seem to have failed to put to rest so far.

Paths are just far more complicated to get right, especially on a case-insensitive file system like that of Windows. The [normalization](https://blogs.msdn.microsoft.com/jeremykuhne/2016/04/21/path-normalization/) rules are even more complicated. I'd like to err on the safe side by avoiding the mess altogether, especially if it's not needed for the problem at hand.

seesharper: Okay, let us focus on getting this as correct as possible.

My suggestion is that we include the path to the script as part of the hash. The only thing that will happen if the user moves the script to another location is that we get a cache miss. No big issue in my opinion.  It might not be ideal, but at least we reduce the chance of formatting hard drives :)

When it comes to cache invalidation we must include the `#load`ed scripts as well. Without that this is not going to work well at all :) It is very common to have more than one script.

Maybe we could make the `ProjectFile` class `IEquatable<T>`  somehow so that we can easily determine if dependencies has changed or if we actually have a deterministic set of dependencies. That will come in handy for another type of cache (another PR 😃) where we can use this information to determine if we need to do a `dotnet restore`. The scenario here is that we keep editing a script in say VS Code and as long as we don't change any dependencies, we don't have to do a restore. 

Finally, but not least, we can only cache if we have a deterministic set of dependencies meaning that we cannot have unpinned NuGet package references. If we are unable to cache because of unpinned packages, we should log it as a debug or warning message.

atifaziz: > The only thing that will happen if the user moves the script to another location is that we get a cache miss. 

That's not the only thing that will happen. On Windows, a cache miss will happen on each _case variation_ of the path that the _user types_, e.g. `C:\temp\foo.csx`, `C:\TEMP\foo.csx`, `C:\TEMP\Foo.csx`. [`Path.GetFullPath`](https://docs.microsoft.com/en-us/dotnet/api/system.io.path.getfullpath?view=netcore-2.1#System_IO_Path_GetFullPath_System_String_) won't help because it won't do conversions to the original case of each part of the path.

seesharper: Okay. Could we do `GetFullPath().ToLower()` on Windows then?

seesharper:
@atifaziz I am working on another type of cache and I need to know if a package reference represents a pinned NuGet version. Take a look here for reference. https://docs.microsoft.com/en-us/nuget/reference/package-versioning. I need something like `IsPinned` on the `PackageReference` class. What do you think ? Doable?

seesharper:
@atifaziz yeah, that makes sense. I am just wondering if `1.2.3` also represents a pinned version? Or is it always just the version specified inside square brackets?

atifaziz:
> I am just wondering if `1.2.3` also represents a pinned version? Or is it always just the version specified inside square brackets?

Technically, and per [Version ranges and wildcards](https://docs.microsoft.com/en-us/nuget/reference/package-versioning#version-ranges-and-wildcards), `1.2.3` means any version _x_ ≥ 1.2.3 but I bet most people _understand_ that to mean exactly 1.2.3. People aside, `1.2.3` is a specification of what's the acceptable range but a tool can have a policy that doesn't have to follow that as long as it's not violating the range. It doesn't mean it _must_ install a later version if available, just that it would be acceptable to do so. NuGet used to be over optimistic and install any version after 1.2.3 but this proved to be trouble some and so [it was changed](https://docs.microsoft.com/en-us/nuget/release-notes/nuget-2.8#patch-resolution-for-dependencies):

> ## Patch Resolution for Dependencies
>
> When resolving package dependencies, NuGet has historically implemented a strategy of selecting the lowest major and minor package version which satisfies the dependencies on the package. Unlike the major and minor version, however, the patch version was always resolved to the highest version. Though the behavior was well-intentioned, it created a lack of determinism for installing packages with dependencies.

If the tool is a cache and we assume that consulting a feed would be expensive for cache validation & currency, I would use a policy that interprets `1.2.3` as `[1.2.3]`.

filipw:
thanks for all the comments and discussions and help - this PR has been superseded by #383 🍻

PR #362, opened 2018-10-07
Add missing exit code assertions in ScaffoldingTests

atifaziz:


src/Dotnet.Script.Tests/ScaffoldingTests.cs
```diff
@@ -59,24 +60,27 @@ public void ShouldRunCsxScriptDirectly()
...
+                var (_, initExitCode) = ScriptTestRunner.Default.Execute("init", scriptFolder.Path);
+                Assert.Equal(0, initExitCode);
 
                 var scriptPath = Path.Combine(scriptFolder.Path, "main.csx");
 
                 if (ScriptEnvironment.Default.IsWindows)
                 {
-                    (output, exitCode) = ProcessHelper.RunAndCaptureOutput("cmd.exe", $"/c \"{scriptPath}\"", scriptFolder.Path);
+                    var (output, exitCode) = ProcessHelper.RunAndCaptureOutput("cmd.exe", $"/c \"{scriptPath}\"", scriptFolder.Path);
+                    Assert.Equal(0, exitCode);
```
filipw: I am not sure it will help here because when it fails it will show something like "expected 0, but was -12345" instead of "expected 'hello world', but was '{stack trace}'"

atifaziz: This was more about consistently checking the exit code for success (tests seems to do this arbitrarily for now) and failing as early as possible. I have changed the assertion to use `Assert.True` instead in 38af98de955285764336a0ba5bc12d132eb3875f and dumped the output as the message on failure. Hope this helps.

PR #365, opened 2018-10-08
Fix "reg add" data escaping & quoting issues

atifaziz:
The association of `.csx` files with `dotnet-script` on Windows via `reg add` had a number of escaping and quoting issues of the value data and honestly I'm not sure how this ever worked. It _might_ also explain the sporadic failures of `ShouldRunCsxScriptDirectly` because escaping of command-line arguments is always tricky. In any case, this PR should make the registration and its test more robust.

Before this PR, running:

    reg query HKCU\Software\Classes\dotnetscript /s

shows the quoting issues:

```
HKEY_CURRENT_USER\Software\Classes\dotnetscript\Shell

HKEY_CURRENT_USER\Software\Classes\dotnetscript\Shell\Open

HKEY_CURRENT_USER\Software\Classes\dotnetscript\Shell\Open\Command
    (Default)    REG_EXPAND_SZ    %ProgramFiles%\dotnet\dotnet.exe exec A:\dotnet-script\src\Dotnet.Script\bin\Debug\netcoreapp2.1\dotnet-script.dll %1 -- %*
```

Note that:

1. `%ProgramFiles%\dotnet\dotnet.exe` isn't quoted, which means that if `%ProgramFiles%` expands to `C:\Program Files` then there's a space that should become part of the program path, but it won't since it's unquoted.
2. The path to `dotnet-script.dll` isn't quoted so if there's a space in the installation path, it'll throw things off.
3. The script path, `%1`, is unquoted and will exhibit the same issues as 1 & 2.

For more information, see the following text section “[Shortcut Menu Verbs](https://docs.microsoft.com/en-us/windows/desktop/shell/fa-verbs#shortcut-menu-verbs)” of “[Verbs and File Associations](https://docs.microsoft.com/en-us/windows/desktop/shell/fa-verbs)”:

> The command string typically looks as follows:
>
>     "My Program.exe" "%1"
>
> If any element of the command string contains or might contain spaces, it must be enclosed in quotation marks. Otherwise, if the element contains a space, it will not parse correctly. For instance, **"My Program.exe"** starts the application properly. If you use **My Program.exe** without quotation marks, then the system attempts to launch **My** with **Program.exe** as its first command line argument. You should always use quotation marks with arguments such as **"%1"** that are expanded to strings by the Shell, because you cannot be certain that the string will not contain a space.

After applying this PR, running:

    reg query HKCU\Software\Classes\dotnetscript /s

shows proper quoting:

```
HKEY_CURRENT_USER\Software\Classes\dotnetscript\shell

HKEY_CURRENT_USER\Software\Classes\dotnetscript\shell\open

HKEY_CURRENT_USER\Software\Classes\dotnetscript\shell\open\command
    (Default)    REG_EXPAND_SZ    "%ProgramFiles%\dotnet\dotnet.exe" exec "A:\dotnet-script\src\Dotnet.Script\bin\Debug\netcoreapp2.1\dotnet-script.dll" "%1" -- %*
```

### DIFF
--- a/src/Dotnet.Script.Core/Scaffolder.cs
+++ b/src/Dotnet.Script.Core/Scaffolder.cs
@@ -132,7 +132,7 @@ namespace Dotnet.Script.Core
 
                 if (_commandRunner.Execute("reg", @"query HKCU\Software\classes\dotnetscript") != 0)
                 {
-                    _commandRunner.Execute("reg", $@"add HKCU\Software\Classes\dotnetscript\Shell\Open\Command /f /ve /t REG_EXPAND_SZ /d """"%ProgramFiles%\dotnet\dotnet.exe"" exec {_scriptEnvironment.InstallLocation}\dotnet-script.dll %1 -- %*""");
+                    _commandRunner.Execute("reg", $@"add HKCU\Software\Classes\dotnetscript\Shell\Open\Command /f /ve /t REG_EXPAND_SZ /d ""\""%ProgramFiles%\dotnet\dotnet.exe\"" exec \""{Path.Combine(_scriptEnvironment.InstallLocation, "dotnet-script.dll")}\"" \""%1\"" -- %*""");
                 }
             }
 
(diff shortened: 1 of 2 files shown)


filipw:
thanks!

atifaziz:
> Let's merge this and rebase the other PR's to see if this fixes the sporadically failing test

@seesharper I merged into all open PRs where builds were superficially failing due to `ShouldRunCsxScriptDirectly` and none of the builds of those merges have failed so far so hope this has done the trick.

PR #366, opened 2018-10-08
Added an option "temporaryDirectory" for execution and publishing

fredpointzero:
I added a `temporaryDirectory` option for default and `publish` commands of the tool.

Motivations:
* My projects uses local development nuget packages only available locally, and I want to use the `#r "nuget: <package>, <version>"` syntax to load those packages from my local feed.
  * The `-s` option to add NuGet feed don't work with the `publish` command
  * => To have proper the proper configuration of NuGet, I set the `temporaryDirectory` to a folder that is a child of my solution (that contains a `NuGet.config`file, thus, all generated csproject will inherit those NuGet settings)

* Also, I like the ability to set a custom temporary folder that can be tied to my workspace

### DIFF
--- a/src/Dotnet.Script.Core/Interactive/InteractiveRunner.cs
+++ b/src/Dotnet.Script.Core/Interactive/InteractiveRunner.cs
@@ -22,6 +22,7 @@ namespace Dotnet.Script.Core
         private ScriptState<object> _scriptState;
         private ScriptOptions _scriptOptions;
         private InteractiveScriptGlobals _globals;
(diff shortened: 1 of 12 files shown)


fredpointzero:
Note: It seems that the resolve is possible with `NuGet.Configuration.Settings.LoadDefaultSettings()` (see https://github.com/NuGet/NuGet.Client/blob/dev/src/NuGet.Core/NuGet.Configuration/Settings/Settings.cs)

PR #368, opened 2018-10-09
Generate a NuGet.config file with resolved properties when publishing a script

fredpointzero:
Motivations:
* I want to publish a script file referencing a nuget package in a local feed. (using the `#r "nuget: package, version"` syntax)

Current state:
* The `-s`option to add NuGet feed doesn't work with publish and is not enough: even if it worked, the `nuget restore` of the generated project would pick the package from the local feed and decompress it in the global NuGet cache.

This was avoided by using a `NuGet.config` file in the solution containing the `.csx` script files.

Solution:
* By resolving all NuGet settings affecting the script file and generating a `NuGet.config` file with those properties for the generated csproj.

With such config file, all settings are also applied during the build steps of the script file.


seesharper:
@fredpointzero Rebased. Could you add a test to back up this new functionality ? 😎

fredpointzero:
@seesharper Thanks for the update, I'll add the test.

filipw:
maybe a better solution would be to copy all dependencies to the DLL output folder like we do with an executable? That way, the script would become independent of the nuget cache. With the current solution (both on master and in this PR), once nuget cache is deleted, running from the DLL will fail.

seesharper:
@filipw We actually do copy deps to the output folder to ensure correctness for platform specific libraries 

https://github.com/filipw/dotnet-script/blob/fed459a52878ae4cc45656a113bd5688425e9ce6/src/Dotnet.Script.Core/ScriptPublisher.cs#L135

So it should be possible to run with these deps instead of the ones from the NuGet cache.

fredpointzero:
@seesharper I added the tests as requested, and they are all green.

src/Dotnet.Script.Core/ScriptPublisher.cs
```diff
@@ -97,6 +98,10 @@ public ScriptPublisher(LogFactory logFactory, ScriptEmitter scriptEmitter)
         {
             try
             {
+                var nugetFile = Path.Combine(outputDirectory, "NuGet.config");
```
seesharper: I think that we would be better off copying this file to the temp folder from which we actually do the restore. Today we copy local NuGet.Config file to the temp folder and do the restore. That works fine unless the you have a NuGet.Config file located in a parent folder relative to the script folder. If we could resolve the properties and write that NuGet.Config file to the temp folder, it would also work from the OmniSharp side of things. OmniSharp also does a potential restore through dotnet script to gather the compilation dependencies needed for language services and such. In addition we would like this to work with the REPL as well. Bottom line here is that this needs to be moved closer into the process of actually doing the underlying dotnet restore so that it works in all use cases and not only when publishing

seesharper: Which would mean that this code probably belongs in the `Dotnet.Script.DependencyModel` library. It also pulls in the `NuGet.Configuration` package . @filipw Is that going to give us any issues over at OmniSharp? OmniSharp already uses a handful of other NuGet related packages.

https://github.com/OmniSharp/omnisharp-roslyn/blob/db13beab49001ec95da7786e29a98ece60e064de/build/Packages.props#L58

fredpointzero: It makes sense, in that way all NuGet properties would be available even if the config file is in a prent folder.
It seems also to be very straightforward, as it can be done by only replacing `ScriptProjectProvider.CopyNuGetConfigFile` with the utility I added.

seesharper: Exactly 👍

seesharper: Do that and I'll have another look. 👍😎

filipw: Nuget packages used to cause problems in OmniSharp since OmniSharp ships with an embedded Mono which trims all the unused types to eventually arrive at 5MB or so runtime. The old packages used drag in a ton of dependencies but since v4 this is no longer the case. I would only advise to use 4.8.0 to align exactly with the rest of O#

src/Dotnet.Script.Tests/ScriptPackagesFixture.cs
```diff
@@ -53,10 +53,10 @@ private void BuildScriptPackages()
             }
         }
 
-        private static string GetPathToPackagesFolder()
+        internal static string GetPathToPackagesFolder()
```
seesharper: Awesome. Good to get rid of those crazy hacks to resolve these packages

PR #370, opened 2018-10-15
Feature/version check

seesharper:
Adds support for checking the latest version against the current version when executing `dotnet script --info`. 

It is implemented by querying the GitHub releases and bring back the latest release. We then simply compare what we get from GitHub with the version we are current executing. If they are different, we display enough information so that the user can update `dotnet-script` to the latest version depending on how `dotnet-script` was installed. 

> Note that we don't attempt to determine if the latest version is "greater" than the current version. We simply assume that if they are different, there is a new version available.


src/Dotnet.Script.Core/Versioning/IVersionProvider.cs
```diff
@@ -0,0 +1,22 @@
+using System.Threading.Tasks;
+
+namespace Dotnet.Script.Core.Versioning
+{
+    /// <summary>
+    /// Represents a class that is capable of providing version information.
+    /// </summary>
+    public interface IVersionProvider
```
filipw: I am not sure we need to introduce interface for this (we don't have them for anything else). for mocking you can just make the methods virtual

seesharper: The reason was to be able to apply a decorator for logging and exception handling. We do something similar for IRestorer. Decorators are possible with virtual methods too although it does not look that clean.  Anyway,  no big issue. I can change it to virtual methods if you want 🙂

seesharper: The reason was to be able to apply a decorator for logging and exception handling. We do something similar for IRestorer. Decorators are possible with virtual methods too although it does not look that clean.  Anyway,  no big issue. I can change it to virtual methods if you want 🙂

seesharper: @filipw  I removed the `IEnvironmentReporter` interface as that provides no real value here and to be perfectly honest it was not even implemented by the `EnvironmentReporter`. I kept the `IVersionProvider` not for the purpose of mocking, but for the purpose of being able to separate logging/exception handling which in turn makes it possible to actually test it. Sounds like a fair compromise?

PR #373, opened 2018-10-18
removed and cleaned up some unneeded code

filipw:
- simplified some repeated code
 - removed the registry registration on Windows for now - due to some of the problems we have been experiencing. we'll resurface this feature at a later time


seesharper:
very good 🎂

PR #391, opened 2018-11-06
Cache improvements

seesharper:
This PR improves on a couple of issues related to caching and publishing. 

* Throw a meaningful error message in the case we fail to find a dependency caused by for instance clearing the global NuGet cache. 

* No need to resolve script package references when running from the cache/dll since the code is already compiled into the DLL

* Discovered during these fixes that we don't used package sources during publish which might also help to overcome the issues discussed in #368


seesharper:
I'll merge this then 👍

PR #402, opened 2018-12-11
CompilationDependencyResolver should operate on known list of CSX files

### LINKED ISSUE #279: Avoid the scan-all behavior against *.csx files
I would like to get rid of is the `Directory.GetFiles(targetDirectory, "*.csx", SearchOption.AllDirectories)` search.

There are three aspects to it.
 - in script execution we follow the loads
 - in REPL mode, we don't need to scan for any CSX, we really just want to create a temporary csproj so that the base runtime assemblies for .NET Core are loaded (and then in case new Nuget references are added)
 - for OmniSharp, OmniSharp itself will discover all CSX files for us, so it could just pass a list of those, and we could set up projects accordingly

filipw:
OmniSharp finds the CSX files once, we can avoid the extra scanning.
Also, OmniSharp allows users to exclude paths from scanning and we should respect that (at the moment we will find those CSX files).

fixes #279

### DIFF
--- a/src/Dotnet.Script.DependencyModel/ProjectSystem/ScriptProjectProvider.cs
+++ b/src/Dotnet.Script.DependencyModel/ProjectSystem/ScriptProjectProvider.cs
@@ -74,6 +74,16 @@ namespace Dotnet.Script.DependencyModel.ProjectSystem
 
         public string CreateProject(string targetDirectory, string defaultTargetFramework = "net46", bool enableNuGetScriptReferences = false)
         {
+            return CreateProject(targetDirectory, Directory.GetFiles(targetDirectory, "*.csx", SearchOption.AllDirectories), defaultTargetFramework, enableNuGetScriptReferences);
+        }
+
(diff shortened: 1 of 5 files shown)


src/Dotnet.Script.DependencyModel/ProjectSystem/ScriptProjectProvider.cs
```diff
@@ -74,6 +74,16 @@ private void LogProjectFileInfo(string pathToProjectFile)
 
         public string CreateProject(string targetDirectory, string defaultTargetFramework = "net46", bool enableNuGetScriptReferences = false)
```
seesharper: Should we mark this as obsolete too?

filipw: this is actually used by us at runtime so it's needed

filipw:
> LGTM. Have you tested this from OmniSharp too? 😀

I will after we merge 😄

PR #410, opened 2019-01-31
Updated to Roslyn 3.0.0 beta2 (C# 8)

filipw:
We can release it as a prerelease version since the latest Roslyn is available as beta on Nuget.

### DIFF
--- a/src/Dotnet.Script.Core/ScriptCompiler.cs
+++ b/src/Dotnet.Script.Core/ScriptCompiler.cs
@@ -34,6 +34,12 @@ namespace Dotnet.Script.Core
             {
                 CSharpScript.Create<object>("1", ScriptOptions.Default, typeof(CommandLineScriptGlobals), new InteractiveAssemblyLoader()).RunAsync(new CommandLineScriptGlobals(Console.Out, CSharpObjectFormatter.Instance)).GetAwaiter().GetResult();
             });
+
+            // reset default scripting mode to latest language version to enable C# 8.0 features
+            // this is not needed once Roslyn 3.0 stable ships
+            var csharpScriptCompilerType = typeof(CSharpScript).GetTypeInfo().Assembly.GetType("Microsoft.CodeAnalysis.CSharp.Scripting.CSharpScriptCompiler");
+            var parseOptionsField = csharpScriptCompilerType?.GetField("s_defaultOptions", BindingFlags.Static | BindingFlags.NonPublic);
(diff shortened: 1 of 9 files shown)


filipw:
`#nullable` should work (we can add a unit test for that though). The generated project file doesn't need to be changed (I think) since we don't use it for compilation options, only for references.

There is a new option on the `CSharpCompilationOptions` that control nullabile reference types, and we could set it - for example based on a command line arg? this way we can avoid `#nullable`

seesharper:
The only issue with a command line arg is that we need to decide what to put in the `launch.json` file.
Anyway, it need to be opt-in. Maybe the pragma is good enough for now?

How is this on the OmniSharp side of things?

Does it show squiggles if you use the pragma? `#nullable enable `

seesharper:
Tried with latest OmniSharp

image


What needs to be done to make this work?

seesharper:
Can the language version be set in `omnisharp.json` ?

src/Dotnet.Script.Core/ScriptCompiler.cs
```diff
@@ -34,6 +34,12 @@ static ScriptCompiler()
             {
                 CSharpScript.Create<object>("1", ScriptOptions.Default, typeof(CommandLineScriptGlobals), new InteractiveAssemblyLoader()).RunAsync(new CommandLineScriptGlobals(Console.Out, CSharpObjectFormatter.Instance)).GetAwaiter().GetResult();
             });
+
+            // reset default scripting mode to latest language version to enable C# 8.0 features
+            // this is not needed once Roslyn 3.0 stable ships
+            var csharpScriptCompilerType = typeof(CSharpScript).GetTypeInfo().Assembly.GetType("Microsoft.CodeAnalysis.CSharp.Scripting.CSharpScriptCompiler");
+            var parseOptionsField = csharpScriptCompilerType?.GetField("s_defaultOptions", BindingFlags.Static | BindingFlags.NonPublic);
+            parseOptionsField?.SetValue(null, new CSharpParseOptions(LanguageVersion.CSharp8, kind: SourceCodeKind.Script));
```
seesharper: Does this execute before 
https://github.com/filipw/dotnet-script/blob/cf01f703499220b3bc48ef329d3458dccb4af40f/src/Dotnet.Script.Core/Interactive/InteractiveRunner.cs#L30

Maybe we should add a nullable type test for the REPL too?

filipw: there is no way to make nullable work yet because it only emits and we swallow warnings at the moment. we also don't expose the ability to treat warnings as errors. I will open a separate issue to handle that

seesharper:
LGTM

PR #420, opened 2019-03-04
updated to Roslyn 3.0.0 beta4

### LINKED ISSUE #418: Update to Roslyn 3.0.0-beta4-final
Matches VS2019 RC

filipw:
Fixes #418


filipw:
we could push a beta though. however, stable Roslyn will ship with VS 2019 so should be in April I think

PR #426, opened 2019-03-12
return script return value from InteractiveRunner.Execute

### LINKED ISSUE #425: Accessing return value from InteractiveRunner.Execute?
Is it possible to access the return value from running an input with the `InteractiveRunner.Execute` method?

filipw:
Fixes #425


cgranade:
Thank you so much for the help!

PR #428, opened 2019-03-18
Feature/nuget projectmodel

seesharper:
Adds support for using `NuGet.ProjectModel` rather than `Microsoft.Extensions.DependencyModel` for reading the dependency context from the generated `project.assets.json` file. This change is needed in order to support the upcoming .Net Core 3.0 release. 

Also added code coverage to the tests using `coverlet` as it made the refactoring a lot easier being able to identify dead code related to dependency resolution. 

With this PR merged we can continue to work with .Net Core 3.0 support which should go a lot easier now that we have removed the dependency on `Microsoft.Extensions.DependencyModel`

Other than that, a few cleanups here and there.


src/Dotnet.Script.DependencyModel/Context/ScriptDependencyContextReader.cs
```diff
@@ -0,0 +1,200 @@
...
+        {
+        }
+
+        public ScriptDependencyContext ReadDependencyContext(string pathToAssetsFile)
+        {
+            var lockFile = LockFileUtilities.GetLockFile(pathToAssetsFile, _nuGetLogger);
+
+            // Since we execute "dotnet restore -r [rid]" we get two targets in the lock file.
+            // The second target is the one containing the runtime deps for the given RID.
+            var targetLibraries = lockFile.Targets[1].Libraries;
```
filipw: can we add here some logic that checks the length first? I worry that in some edge cases this might starts throwing out of range exception

seesharper: Sure, no problem. I'll throw a meaningful exception here

Issue #430, opened 2019-03-31
.csx files are not directly executable on windows on new installations.


tomlm:
Documentation states that I should be able to run a csx file directly by invoking the name.

foo.csx arg1 arg2

This doesn't work on new installations

I am guessing this commit disabled it
https://github.com/filipw/dotnet-script/commit/fb71d31a8f02faabb0c0c78e438dd8e7699da59d

Why was it disabled?  Was there an issue with the code?  Is it possible we could enable it again but only with a switch?

tomlm:
What about adding a new command to register, and only list it for windows?
```dotnet script register```
?

**Comment by tomlm** [comment]
For anyone else who is missing this functionality it can be added from the command line using the **reg** command:
```batch
reg add HKCU\Software\classes\.csx /f /ve /t REG_SZ /d dotnetscript
reg add HKCU\Software\Classes\dotnetscript\Shell\Open\Command /f /ve /t REG_EXPAND_SZ /d "\"%ProgramFiles%\dotnet\dotnet.exe\" script \"%%1\" -- %*"
```

This registers .CSX files -> **dotnet script**

seesharper:
Closed by #431

PR #431, opened 2019-03-31
Add register command for .csx file registration

tomlm:
Added command to platform register .csx files using following command:
**```dotnet-script register```**
To resolve bug #430


seesharper:
Thanks 👍

PR #445, opened 2019-04-19
make dotnet build pass on non-Windows

filipw:
`dotnet build` will now work


filipw:
now yes 😀

seesharper:
Could we also enable the tests in the build script so that it runs for all platforms?

https://github.com/filipw/dotnet-script/blob/3240ee1cc20c603610d3ba4e7b61de54174f8ebe/build/Build.csx#L79

PR #447, opened 2019-04-29
Nullable context support at compile time

### LINKED ISSUE #411: Add support for nullable references
After merging #410 we have Roslyn 3.0.0 and C# 8 so we are able to handle nullable references.
The problem is, even if you enable it with `#nullable` in a CSX file, it emits warnings and at the moment we never show any warnings. 

Possible options:

 - we start always showing warnings too (and potentially add an opt out flag to stop showing them). I like this, but it's potentially a breaking change (different behavior)
 - we add an opt in flag to show warnings
 - we add an opt in flag to treat warnings as errors
 - we add an opt in flag for "nullable references" and we implicitly show warnings + treat warnings as errors then. The advantage of that would be that the code wouldn't need to have the `#nullable` pragma then.

filipw:
Fixes #411 

We treat all nullability diagnostics as errors if the nullable analysis is enabled using the pragma (as discussed).
This should be merged only after we release the current master (so that we make a clean release from master, since this PR bring in Roslyn 3.1.0-beta).

### DIFF
--- a/src/Dotnet.Script.Core/Interactive/InteractiveRunner.cs
+++ b/src/Dotnet.Script.Core/Interactive/InteractiveRunner.cs
@@ -116,6 +116,21 @@ namespace Dotnet.Script.Core
                 _globals.Args.Add(arg);
 
             var compilationContext = ScriptCompiler.CreateCompilationContext<object, InteractiveScriptGlobals>(scriptContext);
+            foreach (var warning in compilationContext.Warnings)
+            {
+                Console.WriteHighlighted(warning.ToString());
+            }
+
+            if (compilationContext.Errors.Any())
+            {
+                foreach (var diagnostic in compilationContext.Errors)
+                {
+                    Console.WriteError(diagnostic.ToString());
+                }
+
+                throw new CompilationErrorException("Script compilation failed due to one or more errors.", compilationContext.Errors.ToImmutableArray());
+            }
+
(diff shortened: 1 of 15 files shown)


src/Dotnet.Script.Core/Interactive/InteractiveRunner.cs
```diff
@@ -116,6 +116,21 @@ private async Task RunFirstScript(ScriptContext scriptContext)
                 _globals.Args.Add(arg);
 
             var compilationContext = ScriptCompiler.CreateCompilationContext<object, InteractiveScriptGlobals>(scriptContext);
+            foreach (var warning in compilationContext.Warnings)
```
seesharper: Maybe we could move this into a method in the `ScriptConsole`. Something like `DumpDiagnostics`?
We have basically the same code in the `ScriptEmitter` class

src/Dotnet.Script.Core/ScriptRunner.cs
```diff
@@ -78,20 +80,23 @@ public Task<TReturn> Execute<TReturn>(ScriptContext context)
 
         public virtual Task<TReturn> Execute<TReturn, THost>(ScriptContext context, THost host)
         {
-            try
+            var compilationContext = ScriptCompiler.CreateCompilationContext<TReturn, THost>(context);
+            foreach (var warning in compilationContext.Warnings)
```
seesharper: Same here. Move this into the `ScriptConsole` ?

PR #449, opened 2019-05-13
Fixed bug in ScriptFilesResolver

seesharper:
This fixes a bug that caused the `ScriptFilesResolver` to ignore `csx`files starting with `nuget`.


filipw:
thanks

PR #452, opened 2019-05-14
Fix the NRE when executing a blank file

### LINKED ISSUE #388: Null reference exception when script is "empty"
If the script only contains the NuGet reference and/or usings without any actual code, there is a NRE thrown.

```
System.NullReferenceException: Object reference not set to an instance of an object.
   at Dotnet.Script.Core.ScriptPublisher.CreateScriptAssembly[TReturn,THost](ScriptContext context, String outputDirectory, String assemblyFileName) in /Users/bernhardrichter/GitHub/dotnet-script/src/Dotnet.Script.Core/ScriptPublisher.cs:line 107
   at Dotnet.Script.Core.ScriptPublisher.CreateAssembly[TReturn,THost](ScriptContext context, LogFactory logFactory, String assemblyFileName) in /Users/bernhardrichter/GitHub/dotnet-script/src/Dotnet.Script.Core/ScriptPublisher.cs:line 45
   at Dotnet.Script.Core.Commands.ExecuteScriptCommand.<GetLibrary>g__CreateLibrary|6_0(<>c__DisplayClass6_0& ) in /Users/bernhardrichter/GitHub/dotnet-script/src/Dotnet.Script.Core/Commands/ExecuteScriptCommand.cs:line 67
   at Dotnet.Script.Core.Commands.ExecuteScriptCommand.GetLibrary(ExecuteScriptCommandOptions executeOptions) in /Users/bernhardrichter/GitHub/dotnet-script/src/Dotnet.Script.Core/Commands/ExecuteScriptCommand.cs:line 54
   at Dotnet.Script.Core.Commands.ExecuteScriptCommand.Run[TReturn,THost](ExecuteScriptCommandOptions options) in /Users/bernhardrichter/GitHub/dotnet-script/src/Dotnet.Script.Core/Commands/ExecuteScriptCommand.cs:line 34
   at Dotnet.Script.Program.<>c__DisplayClass4_0.<<Wain>b__8>d.MoveNext() in /Users/bernhardrichter/GitHub/dotnet-script/src/Dotnet.Script/Program.cs:line 222
   at McMaster.Extensions.CommandLineUtils.CommandLineApplication.<>c__DisplayClass126_0.<OnExecute>b__0() in C:\projects\commandlineutils\src\CommandLineUtils\CommandLineApplication.cs:line 505
   at Dotnet.Script.Program.Main(String[] args) in /Users/bernhardrichter/GitHub/dotnet-script/src/Dotnet.Script/Program.cs:line 26
```

TomasHubelbauer:
Closes #388.

In `var emitResult = _scriptEmitter.Emit<TReturn, THost>(context);`, `emitResult.Success` is `true` but both `PeStream` and `RuntimeDependencies` are then `null` when a blank file is fed in.

I avoided this by checking the length of the source text and adding a return value to the publish command. If the length is non-zero, everything proceeds as usual, but if it is zero, the publish command returns `false` and subsequently that is caught in the execute script command which then proceeds ignore the attempt to execute (otherwise it would fail as the script file would not exist on disk since the publish command cannot create it).

I used a `bool` return value, if you prefer me to throw an exception to control the flow (hopefully you don't) or use an enum for the return value, I am happy to adjust.

I didn't test this beyond `dotnet run empty.csx` in the `Dotnet.Script` project directory.

### DIFF
--- a/src/Dotnet.Script.Core/Commands/ExecuteScriptCommand.cs
+++ b/src/Dotnet.Script.Core/Commands/ExecuteScriptCommand.cs
@@ -32,6 +32,11 @@ namespace Dotnet.Script.Core.Commands
             }
 
             var pathToLibrary = GetLibrary(options);
+            if (pathToLibrary == null)
+            {
+                return default(TReturn);
+            }
+
             return await ExecuteLibrary<TReturn>(pathToLibrary, options.Arguments, options.NoCache);
         }
 
@@ -64,7 +69,12 @@ namespace Dotnet.Script.Core.Commands
             string CreateLibrary()
             {
                 var options = new PublishCommandOptions(executeOptions.File,executionCacheFolder, "script", PublishType.Library,executeOptions.OptimizationLevel, executeOptions.PackageSources, null, executeOptions.NoCache);
-                new PublishCommand(_scriptConsole, _logFactory).Execute(options);
+                var success = new PublishCommand(_scriptConsole, _logFactory).Execute(options);
+                if (!success)
+                {
+                    return null;
+                }
+
                 if (hash != null)
                 {
                     File.WriteAllText(Path.Combine(executionCacheFolder, "script.sha256"), hash);
--- a/src/Dotnet.Script.Core/Commands/PublishCommand.cs
+++ b/src/Dotnet.Script.Core/Commands/PublishCommand.cs
@@ -19,7 +19,7 @@ namespace Dotnet.Script.Core.Commands
             _logger = logFactory.CreateLogger<PublishCommand>();
         }
 
-        public void Execute(PublishCommandOptions options)
+        public bool Execute(PublishCommandOptions options)
         {
             var absoluteFilePath = options.File.Path;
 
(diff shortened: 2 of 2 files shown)


filipw:
This is not going to fully fix it though, because it would work for a completely empty script, but it wouldn't work for example if the whole script content would be something like `using System;`. I am a bit reluctant to add in measuring of code length as a validation point, but if we do, we could also do it at the edge and fail fast.

TomasHubelbauer:
I wasn't aware this had multiple causes. In that case, it is probably better to check this by checking if `PeStream` is not `null`. Maybe I can adjust the code to propagate this info up and instead of check the code length like you say just bubble this other check all the way up and bail when it fails.

I agree that I could bail faster if checking file content length closer to the edge, I thought that would be a bit dirty, but this hack is not much cleaner. If we do decide to use `PeStream != null` though, I don't think we can meaningfully hoist that much higher, so I think I would have to relay that value all the way up using return values of the methods participating in the call chain.

What do you think? Should I use `PeStream != null`?

PR #455, opened 2019-05-19
Removed obsolete logging

seesharper:
This PR bump `Microsoft.Extensions.Logging` to the latest preview. They have finally made an option to log to `stderr` rather than always logging to `stdout`. That is the reason we had to implement our own `IConsole` logger before. This simplifies that quite a bit.


seesharper:
Build seems to time out on Windows. Need to look into that

seesharper:
We wait until Ms.Ex.Logging 3.0 goes RTM

PR #458, opened 2019-05-24
Omnisharp compatibility

seesharper:
This PR enables compilation dependencies to be resolved for a `netcoreapp`target even when the host is desktop .Net. This is mainly to restore compatibility with OmniSharp which is a `net472` app.

* If the host is a non-dotnet core app, we default to `net472` as the default framework.
* Some of the utility classes such as `FileUtils`, `ProcessHelper` and friends has been moved to the `Dotnet.Script.Shared.Tests` project so that they can be used from the `Dotnet.Script.Desktop.Tests` project.


src/Dotnet.Script.Core/Internal/StringExtensions.cs
```diff
@@ -1,6 +1,6 @@
 using System;
 
-internal static class StringExtensions
+public static class StringExtensions
```
filipw: public type in an internal namespace? i'd rather use internals visible to

src/Dotnet.Script.DependencyModel/Environment/ScriptEnvironment.cs
```diff
@@ -55,16 +55,6 @@ private ScriptEnvironment()
 
         public bool IsNetCore => _netCoreVersion.Value != DotnetVersion.Unknown;
 
-        public void OverrideTargetFramework(string targetFramework)
```
filipw: the override should stay if you want to have something different than `net472` (which can be the hardcoded default)
for example for OmniSharp `net472` makes sense, but you can use/host this library in a different context too, so therefore it makes sense to provide a workaround

seesharper: I've reintroduced the framework override. 
Note that the override is no longer used in the Desktop tests as it conflicts with the new compilation dependency tests there. Still able to override, it is just that we use the default for the tests now which is `net472`

src/Dotnet.Script.Core/Internal/StringExtensions.cs
seesharper: Added `InternalsVisibleTo`

src/Dotnet.Script.DependencyModel/Environment/ScriptEnvironment.cs
filipw: yep that's fine

PR #466, opened 2019-06-19
Improved support for local NuGet.Config files.

### LINKED ISSUE #463: Passed config file is modified and becomes invalid
dotnet-script myscript --configfile PathToMzConfigFile

The nuget.config gets copied to the temporary project location. But for some reason it is modified, which is from my point of view not ok. For example nuget.org gets automatically added, which I don't want in an enterprise environment. 

Unfortunately even some important parts are completely removed. The whole packageSourceCredentials section gets removed, which makes dotnet-script fail, when accessing the repositories that require those credentials.

Is there any workaround to avoid that modification?

From my point of view the best solution would be if dotnet-script is executed in the context of the script file. Then the nuget logic to find the config in any folder above would also work, which doesn't currently. 
The other solution would be to find the nuget.config in the directories above when creating the temp project, and copy it ***unmodified*** into the temp script project folder.

seesharper:
This PR adds support for a better handling of `NuGet.Config`files and fixes #463 .

We used to make an attempt to resolve the various NuGet elements/tags from a given script folder. 
This had the drawback of expanding environment variables that could contains sensitive information such as repository credentials. 

As it turned out, `dotnet restore` has a `--configfile` argument that lets us pass the path to the `NuGet.Config` file.  This file may not be in the script folder, but it can also be in a parent folder relative to the script folder. So we use `Settings.GetConfigFilePaths` (NuGet.ProjectModel) for this. 

The first in this list is the "nearest" config file seen from the script folder.

https://github.com/NuGet/NuGet.Client/blob/9dac90978fe16f0b097b4b4b910ffb8dfdc059b6/src/NuGet.Core/NuGet.Configuration/Settings/Settings.cs#L583

Also some general housekeeping like excluding the test scripts from OmniSharp which improves the experience when working with dotnet-script in OmniSharp/VS Code.


src/Dotnet.Script.DependencyModel/Context/DotnetRestorer.cs
```diff
@@ -19,18 +20,18 @@ public DotnetRestorer(CommandRunner commandRunner, LogFactory logFactory)
...
+        public void Restore(ProjectFileInfo projectFileInfo, string[] packageSources)
         {
             var packageSourcesArgument = CreatePackageSourcesArguments();
+            var configFileArgument = CreateConfigFileArgument();
             var runtimeIdentifier = _scriptEnvironment.RuntimeIdentifier;
 
-             _logger.Debug($"Restoring {pathToProjectFile} using the dotnet cli. RuntimeIdentifier : {runtimeIdentifier}");
-            var exitcode = _commandRunner.Execute("dotnet", $"restore \"{pathToProjectFile}\" -r {runtimeIdentifier} {packageSourcesArgument}");
-            //var exitcode = _commandRunner.Execute("dotnet", $"restore \"{pathToProjectFile}\" {packageSourcesArgument}");
+            _logger.Debug($"Restoring {projectFileInfo.Path} using the dotnet cli. RuntimeIdentifier : {runtimeIdentifier}");
```
filipw: let's log here the path to the selected nuget config file

src/Dotnet.Script.DependencyModel/Context/ProfiledRestorer.cs
```diff
@@ -16,11 +17,11 @@ public ProfiledRestorer(IRestorer restorer, LogFactory logFactory)
...
         public bool CanRestore => _restorer.CanRestore;
 
-        public void Restore(string pathToProjectFile, string[] packageSources)
+        public void Restore(ProjectFileInfo projectFileInfo, string[] packageSources)
         {
             var stopwatch = Stopwatch.StartNew();
-            _restorer.Restore(pathToProjectFile, packageSources);
-            _logger.Debug($"Restoring {pathToProjectFile} took {stopwatch.ElapsedMilliseconds}ms");
+            _restorer.Restore(projectFileInfo, packageSources);
+            _logger.Debug($"Restoring {projectFileInfo.Path} took {stopwatch.ElapsedMilliseconds}ms");
```
filipw: or here

src/Dotnet.Script.DependencyModel/Context/DotnetRestorer.cs
seesharper: Added logging

src/Dotnet.Script.DependencyModel/Context/ProfiledRestorer.cs
seesharper: Added in DotnetRestorer

PR #476, opened 2019-07-26
Simplify Dockerfile

GeorgeTsiokos:
- Use the public NuGet release via `dotnet tool install` 
- Use the Azure Container Registry .NET Core 2.1 SDK

### DIFF
--- a/build/Dockerfile
+++ b/build/Dockerfile
@@ -3,4 +3,4 @@ FROM mcr.microsoft.com/dotnet/core/sdk:2.1
 # https://www.nuget.org/packages/dotnet-script/
 RUN dotnet tool install dotnet-script --tool-path /usr/bin
 
-CMD [ "dotnet", "script" ]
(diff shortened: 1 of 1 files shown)


GeorgeTsiokos:
Also uses the newer Azure Container Registry .NET Core 2.1 SDK image URL

filipw:
thanks a lot but I am not convinced about this change. It sort of changes the intent of this `Dockerfile`. it's supposed to contain the dotnet-script version built from the current source instead of the last published stable version.

GeorgeTsiokos:
Personally I’m not a fan of supporting multiple build pipelines, but if that’s the purpose of this Dockerfile then definitely pass on this PR.

In any case please consider publishing an “official” docker image as part of your pipeline so people can discover this great project with a simple ‘docker run dotnet-script’ command from any platform

seesharper:
@filipw Could you re-review 😀according to our internal discussion ?

PR #478, opened 2019-08-01
Handle version number with more than 3 subparts

### LINKED ISSUE #477: Packages not following Major.Minor.Revision versioning cannot be pinned to an exact revision
I'm trying to include some data models from the AndroidPublisher package found at https://www.nuget.org/packages/Google.Apis.AndroidPublisher.v3/

Their versioning scheme seems to have four parts, which means the caching in the dependency resolver fails. Here's the line I'm using to introduce the dependency:

`#r "nuget: Google.Apis.AndroidPublisher.v3, 1.40.2.1648"`

Here's the resulting error:

`The script (myscript.csx) is not cacheable. For caching and optimal performance, 
ensure that the script only contains NuGet references with pinned/exact versions.`

Removing the last part of the version (`#r "nuget: Google.Apis.AndroidPublisher.v3, 1.40.2"`) allows me to cache, but my assumption is that this will ignore any later package releases with version `1.40.2.xxxx` as soon as it has cached something.

This is probably a minor issue affecting only a handful of packages overall, but I thought that it might be worthwhile bringing to your attention. Enabling a way to force caching for non-standard package versions would be my suggestion for working around the issue. Is there a way to do it?

seesharper:
This PR fixes #477 by allowing more than 3 subparts of a version number when determining of we are dealing with a pinned version.

### DIFF
--- a/src/Dotnet.Script.DependencyModel/ProjectSystem/PackageVersion.cs
+++ b/src/Dotnet.Script.DependencyModel/ProjectSystem/PackageVersion.cs
@@ -8,7 +8,7 @@ namespace Dotnet.Script.DependencyModel.ProjectSystem
     /// </summary>
     public class PackageVersion : IEquatable<PackageVersion>
     {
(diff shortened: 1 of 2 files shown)


filipw:
thanks!

PR #482, opened 2019-09-20
Feature/netcoreapp3.0

seesharper:
This PR adds support for running `dotnet-script` on .Net Core 3.0

The most fundamental change is how we resolve compilation and runtime dependencies for NetCoreApp. 
These dependencies used to be reflected in `project.assets.json`, but starting from .Net Core 3.0 these dependencies are added using target files. 
We have reached out the the dotnet/cli team and found a solution that works great for compilation dependencies. 
Take a look here for the details  https://github.com/dotnet/cli/issues/11504
A great thank you goes out to @dsplaisted for guiding us in the right direction. 

We now produce a multi-targetet package that support both netcoreapp2.1 and netcoreapp3.0. It will be installed according to what is available on the target system


src/Dotnet.Script.DependencyModel/Context/ScriptDependencyContextReader.cs
```diff
@@ -48,7 +50,7 @@ public ScriptDependencyContext ReadDependencyContext(string pathToAssetsFile)
             foreach (var targetLibrary in targetLibraries)
             {
                 var scriptDependency = CreateScriptDependency(targetLibrary.Name, targetLibrary.Version.ToString(), packageFolders, packagePathResolver, targetLibrary);
```
filipw: the packageFolders are not used for anything, can we remove them?

src/Dotnet.Script.DependencyModel/ProjectSystem/ScriptProjectProvider.cs
```diff
@@ -124,10 +132,38 @@ public ProjectFile CreateProjectFileFromScriptFiles(string defaultTargetFramewor
                 projectFile.PackageReferences.Add(packageReference);
             }
 
+            // if (defaultTargetFramework == "netcoreapp3.0")
```
filipw: this is a leftover?

src/Dotnet.Script.DependencyModel/ProjectSystem/ScriptProjectProvider.cs
```diff
@@ -124,10 +132,38 @@ public ProjectFile CreateProjectFileFromScriptFiles(string defaultTargetFramewor
...
+            // {
+            //     AddNetCoreAppReference(projectFile);
+            // }
+
             projectFile.TargetFramework = parseresult.TargetFramework ?? defaultTargetFramework;
             return projectFile;
         }
 
+        private void AddNetCoreAppReference(ProjectFile projectFile)
```
filipw: projectFile is never used since the code is commented out, can we remove this?

src/Dotnet.Script.Desktop.Tests/CompilationDepenencyTests.cs
```diff
@@ -14,13 +14,15 @@ public CompilationDependencyTests(ITestOutputHelper testOutputHelper)
             testOutputHelper.Capture();
         }
 
-        [Fact]
-        public void ShouldGetCompilationDependenciesForNetCoreApp2_1()
+        [Theory]
+        [InlineData("netcoreapp2.1")]
+        [InlineData("netcoreapp3.0")]
+        public void ShouldGetCompilationDependenciesForNetCoreApp2_1(string targetFramework)
```
filipw: we should rename this test as it now tests both targets

src/Dotnet.Script.Tests/Dotnet.Script.Tests.csproj
```diff
@@ -1,6 +1,7 @@
 <Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
-    <TargetFrameworks>netcoreapp2.1</TargetFrameworks>
+    <TargetFrameworks>netcoreapp2.1;netcoreapp3.0</TargetFrameworks>
+    <!-- <TargetFramework>netcoreapp3.0</TargetFramework> -->
```
filipw: can we remove this?

seesharper: 👍

src/Dotnet.Script.Desktop.Tests/CompilationDepenencyTests.cs
seesharper: 👍

src/Dotnet.Script.DependencyModel/ProjectSystem/ScriptProjectProvider.cs
seesharper: 👍

PR #485, opened 2019-09-24
Fix compilation dependencies

seesharper:
This PR brings back the `CompilationDepenency` class as it turned out we need more that just the path to the reference assembly in OmnSharp.

In addition this PR makes sure that we create a temporary csproj file per target framework. This is crucial to machines that might have `netcoreapp2.1`  caches (project.assets.json). We could end up running a `netcoreapp3.0` with `netcoreapp2.1` dependencies. Same goes for compilation dependencies.

So this PR also ensures that we get isolated runtime and compilation dependencies per target framework. 
This was not really a problem before as we only had `netcoreapp2.1`. We now have both.

### DIFF
--- a/src/Dotnet.Script.DependencyModel/Compilation/CompilationDependencyResolver.cs
+++ b/src/Dotnet.Script.DependencyModel/Compilation/CompilationDependencyResolver.cs
@@ -1,5 +1,11 @@
+using System;
 using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using Dotnet.Script.DependencyModel.Context;
+using Dotnet.Script.DependencyModel.Environment;
 using Dotnet.Script.DependencyModel.Logging;
+using Dotnet.Script.DependencyModel.Process;
 using Dotnet.Script.DependencyModel.ProjectSystem;
 
 namespace Dotnet.Script.DependencyModel.Compilation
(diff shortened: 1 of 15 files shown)


src/Dotnet.Script.Tests/ScriptExecutionTests.cs
```diff
@@ -31,16 +31,16 @@ public void ShouldExecuteHelloWorld()
         [Fact]
         public void ShouldExecuteScriptWithInlineNugetPackage()
         {
-            var result = ScriptTestRunner.Default.ExecuteFixtureInProcess("InlineNugetPackage");
-            //Assert.Contains("AutoMapper.MapperConfiguration", result.output);
+            var result = ScriptTestRunner.Default.ExecuteFixture("InlineNugetPackage");
```
filipw: what happened with the asserts here?

src/Dotnet.Script.DependencyModel/Compilation/CompilationDependencyResolver.cs
```diff
@@ -1,28 +1,62 @@
...
+            {
+                var compilationDependency = new CompilationDependency(scriptDependency.Name, scriptDependency.Version, scriptDependency.CompileTimeDependencyPaths, scriptDependency.ScriptPaths);
+                result.Add(compilationDependency);
+            }
+
+            // On .Net Core, we need to fetch the compilation references for framework assemblies separately.
+            if (defaultTargetFramework.StartsWith("netcoreapp3", StringComparison.InvariantCultureIgnoreCase))
+            {
+                var compilationreferences = _compilationReferenceReader.Read(projectFileInfo);
+                result.Add(new CompilationDependency("Microsoft.NETCore.App", ScriptEnvironment.Default.NetCoreVersion.Version, compilationreferences.Select(cr => cr.Path).ToArray(), Array.Empty<string>()));
```
filipw: does this actually work? if `ScriptEnvironment.Default.NetCoreVersion.Version` is invoked from omnisharp, it will throw, no?

src/Dotnet.Script.Tests/ScriptExecutionTests.cs
seesharper: Leftover from debugging. Fixing

src/Dotnet.Script.DependencyModel/Compilation/CompilationDependencyResolver.cs
seesharper: No, we don't throw  anymore. It comes back as "unknown". It does not really matter what the version number is. It is not used for anything. Lets just put `3.0` instead 👍

filipw:
we also need a version bump on Dotnet.Script.DependencyModel

seesharper:
Perfect, I'll merge it then 👍

PR #518, opened 2020-02-29
Support custom exe file name with -n

### LINKED ISSUE #517: Allow a name to be passed when creating a self contained executable
Today we currently don't support specifying a name `-n somename` when doing `dotnet script publish`. It also incorrectly default to the name `script` instead of defaulting to the name of the script file.

nodew:
Fixes #517

### DIFF
--- a/src/Dotnet.Script.Core/ScriptPublisher.cs
+++ b/src/Dotnet.Script.Core/ScriptPublisher.cs
@@ -63,7 +63,7 @@ namespace Dotnet.Script.Core
             }
         }
 
-        public void CreateExecutable<TReturn, THost>(ScriptContext context, LogFactory logFactory, string runtimeIdentifier)
+        public void CreateExecutable<TReturn, THost>(ScriptContext context, LogFactory logFactory, string runtimeIdentifier, string executeFileName = null)
         {
             if (runtimeIdentifier == null)
             {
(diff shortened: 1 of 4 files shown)


src/Dotnet.Script.Core/ScriptPublisher.cs
```diff
@@ -63,20 +63,21 @@ public ScriptPublisher(LogFactory logFactory, ScriptEmitter scriptEmitter)
...
             const string AssemblyName = "scriptAssembly";
+            executeFileName = executeFileName ?? Path.GetFileNameWithoutExtension(context.FilePath);
 
-            var tempProjectPath = ScriptProjectProvider.GetPathToProjectFile(Path.GetDirectoryName(context.FilePath), ScriptEnvironment.Default.TargetFramework);
+            var tempProjectPath = ScriptProjectProvider.GetPathToProjectFile(Path.GetDirectoryName(context.FilePath), ScriptEnvironment.Default.TargetFramework, executeFileName);
             var tempProjectDirectory = Path.GetDirectoryName(tempProjectPath);
 
             var scriptAssemblyPath = CreateScriptAssembly<TReturn, THost>(context, tempProjectDirectory, AssemblyName);
-            var projectFile = new ProjectFile(File.ReadAllText(tempProjectPath));
+            var projectFile = new ProjectFile();
```
nodew: What's the purpose of reading temporary project file here? 
Seems that there has been a script.csproj inside the temporary path, but I don't know when and where these files are written.

src/Dotnet.Script.Core/ScriptPublisher.cs
```diff
@@ -63,29 +63,32 @@ public ScriptPublisher(LogFactory logFactory, ScriptEmitter scriptEmitter)
             }
         }
 
-        public void CreateExecutable<TReturn, THost>(ScriptContext context, LogFactory logFactory, string runtimeIdentifier)
+        public void CreateExecutable<TReturn, THost>(ScriptContext context, LogFactory logFactory, string runtimeIdentifier, string executeFileName = null)
```
seesharper: could we name this parameter `executableFileName` ?

src/Dotnet.Script.Core/ScriptPublisher.cs
```diff
@@ -63,29 +63,32 @@ public ScriptPublisher(LogFactory logFactory, ScriptEmitter scriptEmitter)
...
             var tempProjectDirectory = Path.GetDirectoryName(tempProjectPath);
-
+            
             var scriptAssemblyPath = CreateScriptAssembly<TReturn, THost>(context, tempProjectDirectory, AssemblyName);
+           
             var projectFile = new ProjectFile(File.ReadAllText(tempProjectPath));
             projectFile.PackageReferences.Add(new PackageReference("Microsoft.CodeAnalysis.Scripting", ScriptingVersion));
             projectFile.AssemblyReferences.Add(new AssemblyReference(scriptAssemblyPath));
-            projectFile.Save(tempProjectPath);
+            projectFile.Save(renamedProjectPath);
```
filipw: i don't think we should *always* create two different paths. only when the executable file name was overridden, otherwise it should use the old logic to avoid the second unnecessary file?

nodew: From @seesharper, we should default to the name of the script filename rather than `script`, so the overridden happens when there's no `-n` option passed. Only when the script name is `script.csx`,  the two path will be the same.

filipw:
thanks!

PR #521, opened 2020-03-03
Fix: AssemblyResolve handler do not throw Exception

aidmsu:
This PR changes AssemblyResolve handler: if assembly can't be resolved we don't throw an exception but just return null.

**Why?**

For example, we would like to check if our application use Unity. We write:
```
var unityType = Type.GetType("UnityEngine.Application, UnityEngine");
```
We expects that `unityType` equals to null if `UnityEngine.Application` type is not found. But now an exception will be thrown.

Also AssemblyResolve [description](https://docs.microsoft.com/en-us/dotnet/api/system.appdomain.assemblyresolve?view=netframework-4.8#remarks) tells:

> It is the responsibility of the ResolveEventHandler for this event to return the assembly that is specified by the ResolveEventArgs.Name property, or to return null if the assembly is not recognized.


aidmsu:
@seesharper I added the test.

filipw:
thaanks

PR #522, opened 2020-03-04
User spesific temp folder

seesharper:
This PR adds support for user specific temp folders to be using when caching the result of script compilation.

We used to always do `Path.GetTempPath` for this where as this PR now uses `~ /.cache/dotnet-script` on Linux and `~/Library/Caches/dotnet-script` on MacOS.

### DIFF
--- a/src/Dotnet.Script.DependencyModel/ProjectSystem/FileUtils.cs
+++ b/src/Dotnet.Script.DependencyModel/ProjectSystem/FileUtils.cs
@@ -12,7 +12,7 @@ namespace Dotnet.Script.DependencyModel.ProjectSystem
     {
         public static string CreateTempFolder(string targetDirectory, string targetFramework)
         {
-            string pathToProjectDirectory = Path.Combine(GetPathToTempFolder(targetDirectory), targetFramework);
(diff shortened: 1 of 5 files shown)


src/Dotnet.Script.DependencyModel/ProjectSystem/FileUtils.cs
```diff
@@ -41,8 +43,25 @@ public static string GetPathToTempFolder(string targetDirectory)
...
-            var pathToProjectDirectory = Path.Combine(tempDirectory, "scripts", targetDirectoryWithoutRoot);
+            var pathToProjectDirectory = Path.Combine(tempDirectory, "dotnet-script", targetDirectoryWithoutRoot);
             return pathToProjectDirectory;
         }
+
+        public static string GetTempPath()
+        {
+            var userFolder = System.Environment.GetFolderPath(System.Environment.SpecialFolder.UserProfile);
+
+            if (RuntimeInformation.IsOSPlatform(OSPlatform.Linux))
```
canton7: Shouldn't we be using `$XDG_CACHE_HOME`, if set?

filipw: yes, you are right - it will be handled in another PR 👍

PR #545, opened 2020-06-04
Discard script submission result (intentionally)

atifaziz:
A non-game-changing PR to use native syntax for discarding the submission result, which makes it more obvious that it was intended.

### DIFF
--- a/src/Dotnet.Script.Core/ScriptRunner.cs
+++ b/src/Dotnet.Script.Core/ScriptRunner.cs
@@ -47,10 +47,9 @@ namespace Dotnet.Script.Core
             submissionStates[0] = globals;
 
             var resultTask = method.Invoke(null, new[] { submissionStates }) as Task<TReturn>;
-            TReturn returnValue;
             try
             {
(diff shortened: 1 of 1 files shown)


src/Dotnet.Script.Core/ScriptRunner.cs
```diff
@@ -47,10 +47,9 @@ public async Task<TReturn> Execute<TReturn>(string dllPath, IEnumerable<string>
             submissionStates[0] = globals;
 
             var resultTask = method.Invoke(null, new[] { submissionStates }) as Task<TReturn>;
-            TReturn returnValue;
             try
             {
-                returnValue = await resultTask;
+                _ = await resultTask;
```
seesharper, line 52: Can't we just do `await resultTask` here?

atifaziz: Yes, but a discard captures the _intention_ to deliberately ignore the return value _at the call site_. Just awaiting doesn't tell you if the ignorance was an omission or an intention.

Issue #546, opened 2020-06-04
Have "eval" without code arg take code from standard input


atifaziz:
Today, running `dotnet script eval` shows help. I suggest changing the default behaviour to read the code from standard input input. It would enable naturally evaluating a script in pipelines:

    PS> echo 'Console.WriteLine(DateTime.Now)' | dotnet script eval
    PS> 1..10 | % { "Console.WriteLine($_);" } | dotnet run -p .\src\Dotnet.Script -f netcoreapp3.1 -- eval

As well as:

    curl http://example.com/script.csx | dotnet script eval

The above example would make support of running [remote scripts](https://github.com/filipw/dotnet-script/tree/0.52.0#remote-scripts) obsolete.

Since the current default behaviour of showing help is not something someone would be relying on programmatically (e.g. via shell scripts), this change could be considered non-breaking.

An alternative would be that if the code argument given to `dotnet script eval` is exactly a hypen (`-`) then it would mean read code from standard input, but this seems less discoverable and somewhat of a superficial requirement.

atifaziz:
I am happy to submit a PR needed to enable this. I believe the changes would be very straightforward to the following bits:

https://github.com/filipw/dotnet-script/blob/ead32bfc24849376cf9cb07671702fdafd4157d4/src/Dotnet.Script/Program.cs#L86-L90

PR #547, opened 2020-06-05
Have eval read code from "stdin" when code argument is absent

atifaziz:
This is a PR for #546, and while I understand that the idea proposed hasn't been blessed/approved, I thought I'd nonetheless share a PR to demonstrate the minimum changes needed and in case anyone wants to take it for a test drive.

### DIFF
--- a/src/Dotnet.Script/Program.cs
+++ b/src/Dotnet.Script/Program.cs
@@ -83,7 +83,20 @@ namespace Dotnet.Script
                 c.HelpOption(helpOptionTemplate);
                 c.OnExecute(async () =>
                 {
-                    var source = string.IsNullOrWhiteSpace(code.Value) ? await Console.In.ReadToEndAsync() : code.Value;
+                    var source = code.Value;
+                    if (string.IsNullOrWhiteSpace(source))
+                    {
+                        if (Console.IsInputRedirected)
+                        {
+                            source = await Console.In.ReadToEndAsync();
+                        }
+                        else
+                        {
+                            c.ShowHelp();
+                            return 0;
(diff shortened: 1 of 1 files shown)


src/Dotnet.Script/Program.cs
```diff
@@ -83,13 +83,9 @@ private static int Wain(string[] args)
                 c.HelpOption(helpOptionTemplate);
                 c.OnExecute(async () =>
                 {
-                    if (string.IsNullOrWhiteSpace(code.Value))
-                    {
-                        c.ShowHelp();
-                        return 0;
-                    }
+                    var source = string.IsNullOrWhiteSpace(code.Value) ? await Console.In.ReadToEndAsync() : code.Value;
```
filipw, line 86: this is a nice idea, but what if the piped input is not there? it just hangs?

atifaziz: Yes, just like `cat` and many other tools that default to receiving content over standard input. That said, I can imagine it being confusing for many. Having the ability to interactively type the input isn't going to be the common case and the experience varies sufficiently (e.g. line editing & EOF) between terminals and platforms that it will only add to the confusion. I have therefore made an update with ee28276fb32ee6d4bfddea24b0442e9166f28360 where [`Console.IsInputRedirected`](https://docs.microsoft.com/en-us/dotnet/api/system.console.isinputredirected) is used to detect if standard input has been redirected and display help otherwise. So now, we are back to the current behaviour where `dotnet script eval` alone will display help, but if there's code provided as an argument or supplied over standard input, then it's executed.

PR #551, opened 2020-06-06
Log when running from cache

atifaziz:
With this PR, I'd like to suggest adding the cache location to the `dotnet script --info` display. It takes out the guesswork at times.

Before this PR, `dotnet script --info` will print:

```
Version             : 0.53.0
Install location    : A:\dotnet-script\src\Dotnet.Script\bin\Debug\netcoreapp3.1
Target framework    : netcoreapp3.1
.NET Core version   : 3.1.4
Platform identifier : win
```

After this PR, `dotnet script --info` will print:

```
Version             : 0.53.0
Install location    : A:\dotnet-script\src\Dotnet.Script\bin\Debug\netcoreapp3.1
Target framework    : netcoreapp3.1
.NET Core version   : 3.1.4
Platform identifier : win
Scripts temp path   : C:\Users\johndoe\AppData\Local\Temp\dotnet-script
```

### DIFF
--- a/src/Dotnet.Script.Core/Commands/ExecuteScriptCommand.cs
+++ b/src/Dotnet.Script.Core/Commands/ExecuteScriptCommand.cs
@@ -53,6 +53,7 @@ namespace Dotnet.Script.Core.Commands
                 && TryGetHash(executionCacheFolder, out var cachedHash)
                 && string.Equals(hash, cachedHash))
             {
+                _logger.Debug($"Using cached compilation: " + pathToLibrary);
                 return pathToLibrary;
             }
 
--- a/src/Dotnet.Script.Core/Versioning/EnvironmentReporter.cs
+++ b/src/Dotnet.Script.Core/Versioning/EnvironmentReporter.cs
@@ -1,10 +1,12 @@
 ﻿using System.Text;
 using System.Threading.Tasks;
+using Dotnet.Script.DependencyModel.ProjectSystem;
 using Dotnet.Script.DependencyModel.Environment;
 using Dotnet.Script.DependencyModel.Logging;
 
 namespace Dotnet.Script.Core.Versioning
 {
+
     /// <summary>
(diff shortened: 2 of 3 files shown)


filipw:
Thanks for the PR, however, I don't think this is needed. 
Cache path is an internal implementation detail so it shouldn't show up with the main info, just like nuget restore path doesn't show up on `dotnet --info`. IMHO it's enough that it shows up in debug output and is documented in the documentation.

atifaziz:
`dotnet --info` does not show the NuGet or package restore path because those are SDK-specific details. You can get to some of that info through more specific _sub-commands_ like `dotnet nuget locals --list all`.

Also, `dotnet script --info` already does far more than `dotnet --info`, like checking for the latest version and providing instructions for updating so it seems to have deviated from its role model anyway.

> it shows up in debug output

Fair point, but the problem is that the caching information is only emitted when a build takes place (when cache is missed). On subsequent runs with a cache hit, there is no way to discover or get back that information without changes the script source to cause an invalidation. All I see is:

```
PS> dotnet script -d test.csx
dbug: Dotnet.Script.DependencyModel.ProjectSystem.ScriptParser[0]
      Parsing A:\dotnet-script\bar.csx
Hello!
```

 I agree it's an internal implementation detail but it's a substantial performance feature that leaks into the environment and so it's information that can potentially help with diagnostics. While the temp path [is documented now](https://github.com/filipw/dotnet-script/pull/550), 😉 you still have to go looking for it and the [`README.md` in the `master` branch](https://github.com/filipw/dotnet-script/blob/master/README.md), which by the way, won't necessarily reflect your deployed version. If not, you have to go looking for the `README.md` corresponding to [the version you're using](https://github.com/filipw/dotnet-script/releases), but then that may not always have the info.

filipw:
I think the best solution is to just add "running from cache... {path}" to the debug output

atifaziz:
Okay, I can compromise with that. See 8c67160d3713b0bfa74d65d0593fbda5f79544f3.

PR #552, opened 2020-06-08
Add aggressive caching via C# minification

atifaziz:
With this PR, I'd like to suggest a more aggressive caching strategy where the hash is built on _minified_ C# sources. Minification (not uglification) removes extraneous whitespace, comments and region markers from the source code. Since these cannot affect a program's behaviour, indentation/whitespace styling changes and add/removing/updating comments/regions won't constitute as a change that invalidates a cached execution. Otherwise and presently, even a change of a single space to a script or its `#load`-ed parts will generate a new hash and lead to a cache miss.

The PR uses [C# Minifier](https://github.com/atifaziz/CSharpMinifier) for minification. It's been released since a year and battle/field-tested in production against thousands of C# programs. C# Minifier is a hand-authored lexer written, ground-up, from the [C# specification](https://github.com/dotnet/csharplang/blob/892af9016b3317a8fae12d195014dc38ba51cf16/spec/lexical-structure.md). It does not use Roslyn so it is extremely lightweight to use and in processing. It only needs to understand the lexical structure and practically makes no heap allocations (no syntax trees). As a result, there should be no noticeable impact on performance. If you're curious, you can [try C# Minifier online and in your browser](https://atifaziz.github.io/CSharpMinifierDemo/) (via Blazor WebAssembly). It shows changes in compression and hash computation in real-time.

This PR maintains 100% backward compatibility, including (with painstaking care) at the public API surface level. It introduces a new `enum` called `CommandScriptCompilationCacheLevel`:

```c#
public enum CommandScriptCompilationCacheLevel { Default, None, Simple, Aggressive }
```

`Simple` represents the present behaviour where any physical change to the source will miss the cache. `Aggressive` is the goal of this PR that uses minification. `Default` is defined to be `Aggressive`. This provides a risk-free approach for the future where should you ever choose to revert, you can simply redirect `Default` to `Simple`. The interpretation of `Default` is done at run-time. If it was done at the level of the `enum` member (e.g. `Default = Aggressive`) then a change to the default level could cause binary breaking changes.

I've marked constructors taking a Boolean for caching as obsolete. Since this project is still following the 0.x lineage, we can drop those constructors now or in the future. If you think cache levels is an overkill then we can drop it and maintain just a Boolean flag for caching. Most of the changes in this PR are to maintain the public API compatibility. The actual change to enable caching over minified sources is tiny and limited to the `TryCreateHash` method of [`ExecuteScriptCommand`](https://github.com/filipw/dotnet-script/pull/552/files#diff-89e9a03adb6336108d6faab08143a23f).

### DIFF
--- a/src/Dotnet.Script.Core/Commands/CommandScriptCompilationCacheLevel.cs
+++ b/src/Dotnet.Script.Core/Commands/CommandScriptCompilationCacheLevel.cs
@@ -1,0 +1,24 @@
+namespace Dotnet.Script.Core.Commands
+{
+    public enum CommandScriptCompilationCacheLevel
+    {
+        Default,
+        None,
+        Simple,
+        Aggressive,
+    }
+
+    public static class CommandScriptCompilationCacheLevelDefault
+    {
+        public static CommandScriptCompilationCacheLevel Value = CommandScriptCompilationCacheLevel.Aggressive;
+
+        public static CommandScriptCompilationCacheLevel Normalize(this CommandScriptCompilationCacheLevel level) =>
+            level == CommandScriptCompilationCacheLevel.Default ? Value : level;
+    }
+
+    internal static class CommandScriptCompilationCacheLevelConverter
+    {
+        public static CommandScriptCompilationCacheLevel ToCommandScriptCompilationCacheLevel(this bool value) =>
+            value ? CommandScriptCompilationCacheLevel.None : CommandScriptCompilationCacheLevel.Default;
+    }
+}
--- a/src/Dotnet.Script.Core/Commands/ExecuteCodeCommandOptions.cs
+++ b/src/Dotnet.Script.Core/Commands/ExecuteCodeCommandOptions.cs
@@ -1,16 +1,21 @@
+using System;
 using Microsoft.CodeAnalysis;
 
 namespace Dotnet.Script.Core.Commands
 {
     public class ExecuteCodeCommandOptions
     {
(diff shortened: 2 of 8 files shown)


filipw:
thanks a lot for the PR - however we have decided that it is not a functionality that we need at the moment - we'd like to stay dependency free on the critical path of script execution, even at the expense of some unnecessary cache-misses.
I find the minifier very interesting though 🙂

filipw:
I forgot to add that we are considering a plugin system that might potentially be an interesting avenue for this feature.

atifaziz:
> however we have decided that it is not a functionality that we need at the moment - we'd like to stay dependency free on the critical path of script execution, even at the expense of some unnecessary cache-misses.

Thanks for coming back to me on this PR. I hope you don't take it the wrong way if I express some disappointment about the "we" bits reading like a [Jedi council](https://static.wikia.nocookie.net/starwars/images/8/8c/Jedi_Council_RotS.png/revision/latest?cb=20130202031305) making decisions behind closed doors and just passing down a blunt note of the last and final call. It seems counter-intuitive to the collaboration model of open source projects where you'd expect some discussion in the open. That said, I didn't check in advance if something like this would be welcome so I accept your decision.

> I find the minifier very interesting though 🙂

Thanks! 🙂

> I forgot to add that we are considering a plugin system that might potentially be an interesting avenue for this feature.

I see, but how does a plug-in system alleviate any concerns about dependencies being on the critical path of script execution?

PR #573, opened 2020-08-11
Ignore global.json in script folder

### LINKED ISSUE #571: "global.json" interferes with "dotnet restore" of script
If you try to run a C# script from a folder that contains a `global.json` to pin to a .NET Core SDK version then `dotnet restore` that's executed by `dotnet-script` may fail (depending on the SDK version) to restore packages.

PowerShell steps to reproduce:

1. Create a blank directory and set it as the current directory
2. Create a `global.json` file that pins to .NET Core SDK 3.0.100:
   ```
   Set-Content global.json '{ "sdk": { "version": "3.0.100" } }'
   ```
3. Evaluate a script, e.g.:
   ```
   dotnet script eval //
   ```

At the end of step 3, you should get an error along the lines of:

```
Unable to restore packages from 'C:\Users\johndoe\AppData\Local\Temp\dotnet-script\A\temp\Eval\interactive\interactive\netcoreapp3.1\script.csproj'. Make sure that all script files contains valid NuGet references
```

Evaluating a script that's just a comment, as in the last step above, is the simplest way to reproduce this error but the issue applies equally to running a script:

```
❯ Set-Content global.json '{ "sdk": { "version": "3.0.100" } }' # pin SDK
❯ Set-Content test.csx $null # create a blank script file
❯ dotnet script test.csx
Unable to restore packages from 'C:\Users\johndoe\AppData\Local\Temp\dotnet-script\A\temp\netcoreapp3.1\script.csproj'. Make sure that all script files contains valid NuGet references
```

seesharper:
This PR fixes a bug that comes into play if the user has a `global.json` file located in the script folder that is incompatible with `dotnet-script`. We are pretty much tied to the 3.1 sdk since `dotnet-script` is a `netcoreapp3.1`. I have seen this bug in the past also when trying to execute a script in a folder that resolves the "wrong" sdk through a `global.json` file. 

The solution to this is to execute `dotnet restore` with the temp folder as the working directory. By temp folder I mean the folder where the generated `csproj` file is located. 

Also added a `global.json` file to this repo as it made local testing a little bit difficult now having the .Net 5 prerelease SDK's installed. The pinned version is the same 3.1 sdk we install on the build servers. 

Finally also added a test that verifies that a script can run even if we have an "invalid" `global.json` file 

This should fix #571 👍


src/Dotnet.Script.Tests/ScriptExecutionTests.cs
```diff
@@ -470,7 +479,7 @@ private static string CreateTestScript(string scriptFolder)
 
         private static void CreateTestPackage(string packageLibraryFolder)
         {
-            ProcessHelper.RunAndCaptureOutput("dotnet", "new classlib -n NuGetConfigTestLibrary", packageLibraryFolder);
+            ProcessHelper.RunAndCaptureOutput("dotnet", "new classlib -n NuGetConfigTestLibrary -f netstandard2.0", packageLibraryFolder);
```
seesharper, line 482: This is needed if you have the .Net 5 SDK installed since that version no longer defaults to `netstandard2.0`

global.json
```diff
@@ -0,0 +1,5 @@
+{
+  "sdk": {
+    "version": "3.1.102"
```
filipw, line 3: you can add here "rollForward":"latestFeature" to avoid explicit install

seesharper:
Great. Added `rollForward":"latestFeature` . Merging this then 👍😀

PR #587, opened 2020-11-02
Support remote scripts with empty/null Content Type

### LINKED ISSUE #586: Remote script file cannot execute if ContentType is missing
If a remote script file has no content type, an exception is thrown while downloading.

image

AndrewMJordan:
Fixes #586 by properly accounting for null ContentType.


filipw (approved):
thanks!

PR #589, opened 2020-11-10
Support for .Net 5

seesharper:
This PR adds support for .Net 5.

The most significant changes are related to publishing scripts.

we use to do 
```
var path = Path.GetDirectoryName(Assembly.GetEntryAssembly().Location);
```

Trying to do this in a single file app now yields

```
warning IL3000: 'System.Reflection.Assembly.Location' always returns an empty string for assemblies embedded in a single-file app.
```

So instead we load the `scriptAssembly.dll` by its name instead. This also resulted in a small change where we give the emitted assembly a name so that it can be loaded by name.


filipw (approved):

filipw:
🥇 🔥 🎆

PR #591, opened 2020-11-11
fixed CompilationDependencyResolver for .NET 5.0

filipw:
also renamed the compilation dependency name to avoid confusion that this is a real package
I only bumped the version of that project - we don't need to release everything, I think


seesharper (approved):
🎂

PR #595, opened 2020-11-17
updated to Roslyn 3.8.0

filipw:
stable is now out


seesharper (approved):
Awaesome. While you're at it could you also update this?

https://github.com/filipw/dotnet-script/blob/35e1d7ee577b0572d715291404493aec70fbdd9a/src/Dotnet.Script.Core/ScriptPublisher.cs#L16

seesharper (approved):
Perfect 👍

PR #607, opened 2021-03-08
Update Microsoft.Extensions.Logging.Console to 5.0.0

Lucina:
I have a NuGet dependency that uses Microsoft.Extensions.Logging.Abstractions 5.0.0, scripts using it won't compile due to Dotnet.Script's version differing. 5.0.0 is the latest stable version of the logging packages.

### DIFF
--- a/src/Dotnet.Script/Dotnet.Script.csproj
+++ b/src/Dotnet.Script/Dotnet.Script.csproj
@@ -24,7 +24,8 @@
     <ItemGroup>
         <PackageReference Include="Microsoft.CodeAnalysis.CSharp" Version="3.8.0" />
         <PackageReference Include="McMaster.Extensions.CommandLineUtils" Version="2.2.5" />
-        <PackageReference Include="Microsoft.Extensions.Logging.Console" Version="5.0.0" />
+        <PackageReference Include="Microsoft.Extensions.Logging.Console" Version="5.0.0" Condition="'$(TargetFramework)' == 'net5.0'"/>
+        <PackageReference Include="Microsoft.Extensions.Logging.Console" Version="3.1.12" Condition="'$(TargetFramework)' != 'net5.0'"/>
     </ItemGroup>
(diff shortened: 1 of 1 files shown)


Lucina:
That makes sense. This second commit conditionally uses 5.0.0 for net5.0 and uses 3.1.12 (update from 3.1.6) for the rest.

seesharper:
I am not completely sure if this is actually necessary. The `Microsoft.Extensions.Logging.Abstractions`package is not a SDK dependency and the 5.0 version is compatible with `netstandard2.0` so there should be no problem using the 5.0 version with a `netcoreapp3.1` app. Also we still support netcoreapp2.1 which also can consume this package. It should also be noted that the `Microsoft.Extensions.Logging.Abstractions` is only used in the the `dotnet-script` executable and is not shared with OmniSharp. Just want to make sure that this is actually needed before we start to add conditional package references. It is sort of a potential rabbithole 😀

Lucina:
I did notice those TFMs explicitly mentioned in the package dependency list for 5.0.0, but after the reply I thought it might be better to just leave the version as-is but at least update the version for net5.0 just for numerical alignment and to get my scripts to work. My scripts worked on the three builds Dotnet.Script produces using 5.0.0 and without the conditional references, so I'm reverting the conditional commit.

toras9000:
I think this update will resolve the Microsoft.Extensions conflict problem when using the net5.0 package.

It can be reproduced with this short script, but EFCore 5.0 is currently unavailable.
(I know 3.1 is available, but 5.0 is preferable.)
```
#r "nuget: Microsoft.EntityFrameworkCore, 5.0.5"
class MyDbContext : Microsoft.EntityFrameworkCore.DbContext{}
new MyDbContext();
```

I don't know how it behaves when combined with older runtimes, but I think it would be useful to update it if it doesn't cause any problems.

seesharper:
Fixed by #614

Issue #613, opened 2021-04-19
FileLoadException when running C# script


badclone:
I wrote the following C# script (HelloWorld.csx file):

```
#! "netcoreapp3.1"
#r "nuget: System.Text.Encoding.CodePages, 5.0.0"

public class Script
{
    public static void Run()
    {
        try
        {
            System.Text.Encoding.RegisterProvider(System.Text.CodePagesEncodingProvider.Instance);
        }
        catch (Exception ex)
        {
            System.Console.WriteLine(ex.ToString());
        }
    }
}

Script.Run();
```

I'm using dotnet-script (version 1.0.1) and .NET Runtime (version 3.1.14)

When executing this script, I get the following error. Any idea why?

`dotnet script HelloWorld.csx`

> System.IO.FileLoadException: Could not load file or assembly 'System.Text.Encoding.CodePages, Version=5.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a'. Could not find or load a specific file. (0x80131621)
> File name: 'System.Text.Encoding.CodePages, Version=5.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a'
>  ---> System.IO.FileLoadException: Could not load file or assembly 'System.Text.Encoding.CodePages, Version=5.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a'.
>    at System.Runtime.Loader.AssemblyLoadContext.LoadFromPath(IntPtr ptrNativeAssemblyLoadContext, String ilPath, String niPath, ObjectHandleOnStack retAssembly)
>    at System.Runtime.Loader.AssemblyLoadContext.LoadFromAssemblyPath(String assemblyPath)
>    at System.Reflection.Assembly.LoadFrom(String assemblyFile)
>    at System.Reflection.Assembly.LoadFromResolveHandler(Object sender, ResolveEventArgs args)
>    at System.Runtime.Loader.AssemblyLoadContext.InvokeResolveEvent(ResolveEventHandler eventHandler, RuntimeAssembly assembly, String name)
>    at System.Runtime.Loader.AssemblyLoadContext.OnAssemblyResolve(RuntimeAssembly assembly, String assemblyFullName)
>    at Submission#0.Script.Run()
>    at Submission#0.<<Initialize>>d__0.MoveNext() in D:\scripts\HelloWorld.csx:line 18
> --- End of stack trace from previous location where exception was thrown ---
[...]

PR #614, opened 2021-04-20
Bump Microsoft packages to 5.0.0

seesharper:
This PR bumps the following packages to 5.0.0

* System.Configuration.ConfigurationManager
* System.Text.Encoding.CodePages
* Microsoft.Extensions.Logging.Console

These packages does not affect OmniSharp since they are only used in the `Dotnet.Script` and `Dotnet.Script.Core` projects.

Also added a test for #613 which is fixed by this PR.

All three packages are `netstandard2.0` compliant and can be used from all target frameworks (`netcoreapp2.1`, `netcoreapp3.1` and `net5.0`) 

Finally the tests for `netcoreapp2.1` and `netcoreapp3.1` has been enabled again. We haven't tested those since we added support for `net5.0` 😎


filipw (approved):
thank you!!

PR #616, opened 2021-04-25
fixed publish for net5.0

dangar61:


AndreSteenveld:
I'm getting a similar issue with .NET6 when publishing:

```bash
$ dotnet script --verbosity d publish --configuration Release --output ./ ./project.csx

# ... snip ...
dbug: Dotnet.Script.DependencyModel.Process.CommandRunner[0]
      /usr/share/dotnet/sdk/6.0.201/Sdks/Microsoft.NET.Sdk/targets/Microsoft.NET.Sdk.targets(1110,5): warning NETSDK1179: One of '--self-contained' or '--no-self-contained' options are required when '--runtime' is used. [ ... ]
# ... snip ...
```

It does build successfully but the resulting binary doesn't work at all, I'm guessing this PR would fix that?

PR #628, opened 2021-08-04
Don't stop at entry when debugging in VS Code

seesharper:
This PR changes the default for `stopAtEntry` from `true` to `false`. Recent versions of the debugger shipped with VS Code seems to respect this flag which it did not when we initially created this. The behaviour is unfortunate since the user might think te breakpoint is not being hit. Changing this to `false` ensures that the debugger does not pause until it sees an actual breakpoint.


filipw (approved):

filipw:
thanks!

PR #629, opened 2021-08-17
Test issues when running from a folder containing a space character in its name

hrumhurum:
When the DotNet.Script solution resides in a folder containing a space character in its name, an avalanche of test failures occurs due to the absence of escaping of command line arguments.

The enclosed pull request fixes all those issues.

Besides tests, there was a similar issue in `DotNetRestorer` class that was fixed as well by employing a batte-tested string escaping algorithm from [Gapotchenko.FX.Diagnostics.CommandLine](https://github.com/gapotchenko/Gapotchenko.FX/tree/master/Source/Gapotchenko.FX.Diagnostics.CommandLine) package.

An intricate detail of NuGet's `-s` argument is that it may end with `\` character (e.g. the last separator in a folder path), so such a naive escaping approach as `$"-s \"{s}\""` would not work here, as it will cause an unanticipated escape of the escape character `"`.  Hence the usage of a proper escape function here to isolate the customer-supplied data.

### DIFF
--- a/src/Dotnet.Script.DependencyModel/Context/DotnetRestorer.cs
+++ b/src/Dotnet.Script.DependencyModel/Context/DotnetRestorer.cs
@@ -2,6 +2,7 @@
 using Dotnet.Script.DependencyModel.Logging;
 using Dotnet.Script.DependencyModel.Process;
 using Dotnet.Script.DependencyModel.ProjectSystem;
+using Gapotchenko.FX.Diagnostics;
 using System;
 using System.IO;
 using System.Linq;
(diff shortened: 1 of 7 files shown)


filipw:
Thanks for catching this!

> Besides tests, there was a similar issue in `DotNetRestorer` class that was fixed as well by employing a batte-tested string escaping algorithm from [Gapotchenko.FX.Diagnostics.CommandLine](https://github.com/gapotchenko/Gapotchenko.FX/tree/master/Source/Gapotchenko.FX.Diagnostics.CommandLine) package.

One not ehere, I would prefer to inline the escaping logic because the dependency model package [ships with OmniSharp](https://github.com/OmniSharp/omnisharp-roslyn/blob/master/build/Packages.props#L17) and we intentionally do not want to create extra dependencies for OmniSharp.

hrumhurum:
>I would prefer to inline the escaping logic because the dependency model package ships with OmniSharp and we intentionally do not want to create extra dependencies for OmniSharp.

Done. The corresponding code is now inlined.

seesharper (approved):

filipw:
thanks!

Issue #598, opened 2020-12-08
Memory leak when executing scripts in a loop dynamically in process


gimatech:
Hello,

Calling the ScriptRunner Execute method in a loop leads to a memory leak due to the fact the event AppDomain.CurrentDomain.AssemblyResolve  (subscribed in [ScriptCompiler](https://github.com/filipw/dotnet-script/blob/b5ce1af8b65bd2537b8b960b7852a28ab1475d61/src/Dotnet.Script.Core/ScriptCompiler.cs#L124) is never unsubscribed.

I didn't find a workaround for this
![Memory Dump - Retained ref](https://user-images.githubusercontent.com/8150414/101521252-dd3b0b80-3985-11eb-9f93-8b55ef2f4dea.png)

stevehansen:
Fix seems easy enough (I started my fork for the changes https://github.com/stevehansen/dotnet-script/commit/6f295e2ef4c459308bd1fbd7637e9686fad50252) but we are also seeing if it is possible to use the .NET core assembly unloading https://docs.microsoft.com/en-us/dotnet/standard/assembly/unloadability

PR #630, opened 2021-08-17
Memory leak in AppDomain.AssemblyResolve event subscription

hrumhurum:
There is a memory leak in `AppDomain.AssemblyResolve` event subscription that manifests itself on subsequent script executions.

A similar issue is reported in https://github.com/filipw/dotnet-script/issues/598.

The proposed changes reliably fix the issue.


seesharper (approved):
LTGM 👍. This is something I totally forgot about. Thanks 😊

PR #661, opened 2022-02-09
Allow arrow keys and history to work on linux based systems

cyperous:
I've attempted to test this on Linux (archlinux) and macOS with both dotnet 3.1 and 6.0, and I'm able to use arrow keys to navigate the next.  I've also added history functionality to work in these environments.


cyperous:
Even though this code works in practice it appears the testing framework is preventing it from being testable.  I'll see if I can find a solution.

PR #667, opened 2022-03-07
small improvements in the code

filipw:
I was working on something else and then got sidetracked by the lack of consistency in using declarations, deconstructions etc. so I unified most of it to remove most warnings and compiler suggestions


seesharper (approved):
Great👍. Code needs some cleanup once in a while 👍

PR #682, opened 2022-09-16
Added support for net7.0

seesharper:
This PR adds support for the upcoming `net7.0` which is currently in RC1.

## Removed support for `netcoreapp3.1`

`netcoreapp3.1` reaches EoS (End of Support) on December 13, 2022 and we now support `net5.0`, `net6.0` and `net7.0`.

## Moved build to Github Actions

The build has been moved from Azure Devops to GitHub Actions. The old build was sort of a mess and a lot was cleaned up here.  All jobs now basically have the exact same setup.

## OmniSharp 

`Dotnet.Script.DependencyModel`  and `Dotnet.Script.DependencyModel.Nuget` need to be updated in Omnisharp to support `net7.0`. We used to have a framework check in `CompilationDependencyResolver` which was a leftover from when we supported `netcoreapp2.1`. This check is now removed meaning that with regards to `netcoreapp3.1` which still has a couple of months left, it will still work for resolving compilation dependencies. 

## Versions

The versions has been bumped to 1.4.0 leaving a bit of room if we should need to fix something before `netcoreapp3.1` goes out of support. 

In my opinion we can release this now and get `Dotnet.Script.DependencyModel`  and `Dotnet.Script.DependencyModel.Nuget` updated in Omnisharp in good time before `net7.0` goes RTM.
Hopefully a new version of OmniSharp will be released by then and we avoid resolving to `latest` for users that wants to start using `net7.0` 

## Build Script 

Just a couple of path-related issues where we in the past assumed running the script from the `build` folder.


src/Dotnet.Script/Dotnet.Script.csproj
```diff
@@ -1,10 +1,10 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
     <PropertyGroup>
         <Description>Dotnet CLI tool allowing you to run C# (CSX) scripts.</Description>
-        <VersionPrefix>1.3.1</VersionPrefix>
+        <VersionPrefix>1.4.0</VersionPrefix>
         <Authors>filipw</Authors>
         <PackageId>Dotnet.Script</PackageId>
-        <TargetFrameworks>net6.0;net5.0;netcoreapp3.1</TargetFrameworks>
+        <TargetFrameworks>net7.0;net6.0;net5.0</TargetFrameworks>
```
filipw, line 7: .NET 5.0 is also out of support since May 2022 so we can remove it now

.github/workflows/main.yml
```diff
@@ -0,0 +1,65 @@
...
+  build-linux:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v3
+      - name: Install .Net Core
+        uses: actions/setup-dotnet@v2
+        with:
+          dotnet-version: |
+            5.0.x
```
filipw, line 15: same in CI

filipw (approved):
thanks! let' get rid of 5.0 too

seesharper:
@filipw Removed `net5.0`. 
Could you remove the `Azure Devops` checks?

I don't seem to have the permissions to change the settings 😃

filipw:
done, thanks!!

seesharper:
Great. I'll release this then?. After that Omnisharp needs to be updated with the latest dotnet-script packages. 👍🙂

PR #686, opened 2022-09-25
Don't write diagnostics to stdOut

### LINKED ISSUE #657: Stop compiler warnings being written to stdout
I've written a dotnet script which I'm using as part of a unix pipe chain.

A problem I'm finding is that if I leave some  non-breaking code in the csx file, that throws a warning, e.g.

```
public async Task Foo()
{
    //no code in this method currently uses await, but I can't face refactoring all the 
    //task code out of existence as I'm probably going to use some await code in a minute
}
```
then 
```
warning CS1998: This async method lacks 'await' operators and will run synchronously. Consider using the 'await' operator to await non-blocking API calls, or 'await Task.Run(...)' to do CPU-bound work on a background thread.
```
will be written to StdOut, messing up my unix piping.

is there a way to suppress these warnings, or have them written to StdErr (which is ignored when piping)

seesharper:
This PR fixes #657 by writing warnings from the compiler (Diagnostics) to `stdErr` instead of `stdOut` as we used to. 
As mentioned in the issue this causes "pollution" of the `stdOut` stream and could cause issues when for instance piping the output from `dotnet.script` to another command. 

Tried to make this with as little changes to the tests as possible. We now return a `ProcessResult` with a deconstruct that matches the tuple used in many tests.

### DIFF
--- a/src/Dotnet.Script.Core/ScriptConsole.cs
+++ b/src/Dotnet.Script.Core/ScriptConsole.cs
@@ -36,22 +36,29 @@ namespace Dotnet.Script.Core
             Console.ResetColor();
         }
 
+        public virtual void WriteWarning(string value)
+        {
+            Console.ForegroundColor = ConsoleColor.Yellow;
+            Error.WriteLine(value.TrimEnd(Environment.NewLine.ToCharArray()));
+            Console.ResetColor();
+        }
+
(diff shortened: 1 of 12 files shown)


filipw (approved):
thanks!

PR #687, opened 2022-09-26
Remove Newtonsoft.Json 

### LINKED ISSUE #674: Remove the dependency on Newtonsoft.Json
Having dependency on Newtonsoft.Json cause problems like this: [#663 ](https://github.com/filipw/dotnet-script/issues/663)
Since only two places in whole code use it (Not counting the tests). It will be easy to replace this with native JSON implementation.

[VersionProvider ](https://github.com/filipw/dotnet-script/blob/456414a97956731472eccbc4dcaf1f7e5656de9b/src/Dotnet.Script.Core/Versioning/VersionProvider.cs)

[Scaffolder](https://github.com/filipw/dotnet-script/blob/456414a97956731472eccbc4dcaf1f7e5656de9b/src/Dotnet.Script.Core/Scaffolder.cs)

seesharper:
This PR removes the dependency we have on `Newtonsoft.Json`  and replaces the code with `System.Text.Json`.
`Newtonsoft.Json` has caused quite a few problems over the years since we specify one version in `dotnet-script` while a script might pull in another version.  

`System.Text.Json` is pulled in by default when the host is `net6.0` or `net7.0`. `Dotnet.Script.Core` also support `netstandard2.0` and an conditional package-reference has been added for `System.Text.Json`.

Note: We are not yet completely free from `Newtonsoft.Json` yet since `NuGet.ProjectModel` also has this in its dependency graph, but we have done what we can on the `dotnet-script` side of things and we can close #674 for now.

### DIFF
--- a/src/Dotnet.Script.DependencyModel/Dotnet.Script.DependencyModel.csproj
+++ b/src/Dotnet.Script.DependencyModel/Dotnet.Script.DependencyModel.csproj
@@ -32,7 +32,7 @@
   </ItemGroup>
 
   <ItemGroup>
-    <PackageReference Include="NuGet.ProjectModel" Version="6.3.0" />
+    <PackageReference Include="NuGet.ProjectModel" Version="6.2.0" />
     <PackageReference Include="Microsoft.DotNet.PlatformAbstractions" Version="3.1.6" />
   </ItemGroup>
 
--- a/src/Dotnet.Script.DependencyModel/Dotnet.Script.DependencyModel.csproj
+++ b/src/Dotnet.Script.DependencyModel/Dotnet.Script.DependencyModel.csproj
@@ -32,7 +32,7 @@
   </ItemGroup>
 
   <ItemGroup>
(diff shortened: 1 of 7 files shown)


src/Dotnet.Script.DependencyModel/Dotnet.Script.DependencyModel.csproj
```diff
@@ -32,7 +32,7 @@
   </ItemGroup>
 
   <ItemGroup>
-    <PackageReference Include="NuGet.ProjectModel" Version="5.2.0" />
```
filipw, line 35: omnisharp is still on `6.3.0-rc.128`, let's not use a higher version yet, so I'd suggest 6.2?

filipw (approved):

NiSHoW:
when will this change be released?

seesharper:
We will release a new version just about when .Net 7 goes RTM 👍

PR #690, opened 2022-10-24
removed unnecessary CS8073 warnings

filipw:
SyntaxTrivia and SyntaxToken are non nullable

> /Users/filipw/dev/dotnet-script/src/Dotnet.Script.Core/Internal/PreprocessorLineRewriter.cs(28,17): warning CS8073: The result of the expression is always 'true' since a value of type 'SyntaxTrivia' is never equal to 'null' of type 'SyntaxTrivia?' [/Users/filipw/dev/dotnet-script/src/Dotnet.Script.Core/Dotnet.Script.Core.csproj::TargetFramework=net6.0]
/Users/filipw/dev/dotnet-script/src/Dotnet.Script.Core/Internal/PreprocessorLineRewriter.cs(31,21): warning CS8073: The result of the expression is always 'true' since a value of type 'SyntaxToken' is never equal to 'null' of type 'SyntaxToken?' [/Users/filipw/dev/dotnet-script/src/Dotnet.Script.Core/Dotnet.Script.Core.csproj::TargetFramework=net6.0]

### DIFF
--- a/src/Dotnet.Script.Core/Internal/PreprocessorLineRewriter.cs
+++ b/src/Dotnet.Script.Core/Internal/PreprocessorLineRewriter.cs
@@ -22,13 +22,13 @@ namespace Dotnet.Script
             return HandleSkippedTrivia(base.VisitReferenceDirectiveTrivia(node));
         }
 
-        private SyntaxNode HandleSkippedTrivia(SyntaxNode node)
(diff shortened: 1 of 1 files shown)


seesharper (commented):
Great 👍😃

PR #693, opened 2022-11-09
updates for .NET 7.0

filipw:
note: I did not use the latest 4.4.0 version of Roslyn because it is newer than what is in OmniSharp (it takes them from private feeds)


seesharper (approved):
Great 👍🙂.

PR #694, opened 2022-11-24
Add support for specifying Microsoft.NET.Sdk.Web as the sdk

seesharper:
This PR adds support for specifying an SDK using the `#r` directive.
The feature is restricted to `Microsoft.NET.Sdk.Web` at the moment.

Example 
```csharp
#r "sdk:Microsoft.NET.Sdk.Web"

using Microsoft.AspNetCore.Builder;

var a = WebApplication.Create();
a.MapGet("/", () => "Hello world");
a.Run();
```
This PR also includes a number of updates to NuGet packages referenced by `dotnet-script`. This should minimize the need for `--isolated-load-context`. Also removed a few dependencies that are no longer needed. Some of them was just referenced to make things work back when we were at netcoreapp2.1. 

Other news. Removed support for Chocolatey 😃


README.md
```diff
@@ -8,19 +8,19 @@ Run C# scripts from the .NET CLI, define NuGet packages inline and edit/debug th
...
+| ------------------------------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------- | ----------------------------------- |
+| `dotnet-script` (global tool)         | [![Nuget](http://img.shields.io/nuget/v/dotnet-script.svg?maxAge=10800)](https://www.nuget.org/packages/dotnet-script/)                                             | `net6.0`, `net5.0`, `netcoreapp3.1` |
+| `Dotnet.Script` (CLI as Nuget)        | [![Nuget](http://img.shields.io/nuget/v/dotnet.script.svg?maxAge=10800)](https://www.nuget.org/packages/dotnet.script/)                                             | `net6.0`, `net5.0`, `netcoreapp3.1` |
+| `Dotnet.Script.Core`                  | [![Nuget](http://img.shields.io/nuget/v/Dotnet.Script.Core.svg?maxAge=10800)](https://www.nuget.org/packages/Dotnet.Script.Core/)                                   | `netcoreapp3.1` , `netstandard2.0`  |
```
filipw, line 15: I think this should be marked as `net6.0` not 3.1

src/Dotnet.Script.Tests/ScriptExecutionTests.cs
```diff
@@ -21,8 +21,9 @@ public ScriptExecutionTests(ITestOutputHelper testOutputHelper)
         [Fact]
         public void ShouldExecuteHelloWorld()
         {
-            var (output, _) = ScriptTestRunner.Default.ExecuteFixture("HelloWorld", "--no-cache");
-            Assert.Contains("Hello World", output);
+            ScriptTestRunner.ExecuteFixtureInProcess("HelloWorld", "--no-cache");
+            // var (output, _) = ScriptTestRunner.Default.ExecuteFixture("HelloWorld", "--no-cache");
+            // Assert.Contains("Hello World", output);
```
filipw, line 26: what happened here?

README.md
```diff
@@ -569,6 +565,24 @@ We will also see this when working with scripts in VS Code under the problems pa
...
+#r "sdk:Microsoft.NET.Sdk.Web"
+
+using Microsoft.AspNetCore.Builder;
+
+var a = WebApplication.Create();
+a.MapGet("/", () => "Hello world");
+a.Run();
+```
+
+> Please note the the only SDK currently supported is `Microsoft.NET.Sdk.Web`
```
filipw, line 584: different project SDKs have [different default namespaces](https://learn.microsoft.com/en-us/dotnet/core/project-sdk/overview#implicit-using-directives) which we do not currently do with this implementation. 

we should either have a code path that inserts different default usings based on SDK, or (which I think is fine) we can add as a note here too that this is also not supported for now

filipw (approved):
looks good! I just left a few small comments

src/Dotnet.Script.Tests/ScriptExecutionTests.cs
seesharper: Forgot about that one. Just to be able to test in process during development. Fixing

README.md
seesharper: Yeah, I actually started on this one, but realized that we need to align this with OmniSharp. One way to solve this is to pass the usings to OmniSharp as part of the compilation dependencies. I think we will save that for another PR as this is getting big enough as it is. Creating a comment on this in the code

README.md
seesharper: You´re right. Fixing

filipw (approved):

filipw:
Thanks!

ihqtim:
Great, much appreciated!

PR #699, opened 2023-01-31
allow DOTNET_SCRIPT_CACHE_LOCATION to be relative

### LINKED ISSUE #658: Bug - DOTNET_SCRIPT_CACHE_LOCATION doesn't work with relative path
Because the default cache location for on my linux build server is readonly, I have to set `DOTNET_SCRIPT_CACHE_LOCATION` 

If I use 
```
mkdir _cache
export DOTNET_SCRIPT_CACHE_LOCATION=_cache
dotnet script somescript.csx
```

I get a `project file not found` error as, I think the working directory gets incorrectly combined into a non-existent path (sorry I don't have the example to hand).

I can work around it by writing:

```
mkdir _cache
export DOTNET_SCRIPT_CACHE_LOCATION=$(pwd)/_cache
dotnet script somescript.csx
```

filipw:
If the path retrieved from DOTNET_SCRIPT_CACHE_LOCATION is not absolute, we will make it relative to the current folder.

Fixes #658

### DIFF
--- a/src/Dotnet.Script.DependencyModel/ProjectSystem/FileUtils.cs
+++ b/src/Dotnet.Script.DependencyModel/ProjectSystem/FileUtils.cs
@@ -54,6 +54,7 @@ namespace Dotnet.Script.DependencyModel.ProjectSystem
 
             if (!string.IsNullOrEmpty(cachePath))
             {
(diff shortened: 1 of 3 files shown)


seesharper (approved):
All good 👍😃

PR #724, opened 2023-05-22
Fix greedy parsing of load directive part

### LINKED ISSUE #720: csx parsing problems
I found a few problems with csx parsing:

1. #load directive with next expression start with string
```csharp
#nullable enable
#load "ScriptHelper.csx"

"Hello world".Dump();
```
Produce error:
`Could not find file '/Users/mark/Code/CSharpLab/ScriptHelper.csx"'.`

Although it works:
```csharp
#nullable enable
#load "ScriptHelper.csx"

;
"Hello world".Dump();
```

2.  using statement without parenthesis or a block on top level
```csharp
using var connection = new SqliteConnection($"Data Source={dbPath}");
```
Produce errors:
`error CS0103: The name 'connection' does not exist in the current context`
`error CS1002: ; expected`

Although it works:
```csharp
void Func()
{
  using var connection = new SqliteConnection($"Data Source={dbPath}");
}
```
3.  [Reference to Microsoft.NET.Sdk.Web not work · Issue #716 · dotnet-script/dotnet-script](https://github.com/dotnet-script/dotnet-script/issues/716)

atifaziz:
This PR fixes #720.

It also adds a test to avoid a regression in the future.

### DIFF
--- a/src/Dotnet.Script.DependencyModel/ProjectSystem/ScriptFilesResolver.cs
+++ b/src/Dotnet.Script.DependencyModel/ProjectSystem/ScriptFilesResolver.cs
@@ -61,7 +61,7 @@ namespace Dotnet.Script.DependencyModel.ProjectSystem
 
         private static string[] GetLoadDirectives(string content)
         {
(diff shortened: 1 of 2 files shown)


seesharper (approved):
Thanks. Looks great 😃

PR #725, opened 2023-05-25
Remove hardwired return type in ExecuteScriptCommand

isaacvale:
ExecuteScriptCommand return type should be generic. This PR removes the hardwired return type, pushes it up to Program.cs and passes generic type.

### DIFF
--- a/src/Dotnet.Script.Core/Commands/PublishCommand.cs
+++ b/src/Dotnet.Script.Core/Commands/PublishCommand.cs
@@ -17,6 +17,11 @@ namespace Dotnet.Script.Core.Commands
             _logFactory = logFactory;
         }
 
+        public void Execute(PublishCommandOptions options)
+        { 
+            Execute<int>(options);
+        }
+
         public void Execute<TReturn>(PublishCommandOptions options)
         {
             var absoluteFilePath = options.File.Path;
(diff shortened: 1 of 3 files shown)


filipw:
Can you explain your use case?

seesharper (approved):

src/Dotnet.Script.Core/Commands/PublishCommand.cs
```diff
@@ -17,7 +17,7 @@ public PublishCommand(ScriptConsole scriptConsole, LogFactory logFactory)
             _logFactory = logFactory;
         }
 
-        public void Execute(PublishCommandOptions options)
+        public void Execute<TReturn>(PublishCommandOptions options)
```
filipw, line 20: makes sense thanks. It's technically a breaking change - could you please keep the old one too and add the generic as a new overload?

It does not matter that much but I would like to have backwards compatibility in the public API unless we rev up major version

isaacvale: Yes, you're right, should be an overload, backwards compatibility is very important. Sure, I'll add it

filipw (approved):
thanks!

PR #732, opened 2023-06-21
Add support for .NET Framework

samwebster:


samwebster:
Not sure if any changes are needed to reflect this in the nuget metadata?

seesharper:
I'm afraid that this is not as simple as just adding a new target framework. `dotnet-script` has never supported the full .Net Framework and I doubt that we ever will. We rely heavily on the the Dotnet SDK to make `dotnet-script` work as a global CLI tool. Even if we could make it work for the most simple cases there would be an enormous effort needed to make all tests pass for .Net Framework. If you're looking for a .Net Framework alternative to `dotnet-script`, you might want to check out `csc.exe` which comes bundled with .Net Framework.

filipw:
Thanks a lot, it is intentional not to support .NET Framework, and as @seesharper the support is not so easy.
The Dotnet.Script.Core library is published as netstandard2.0 to NuGet to enable building stuff on top of it, if someone is interested, although that only contains a subset of features.

PR #742, opened 2023-11-13
conditionally reference appropriate version of Microsoft.Extensions.Logging.Console

filipw:
For .NET 8.0 we should bring in the v8 and for .NET 7.0 the v7.

This allows us to remove the isolated load context from the test.


seesharper (approved):
Ah, perfect 👍👌

PR #744, opened 2023-11-14
Updated to Roslyn 4.8.0-3.final

filipw:


filipw:
I think this can be followed by a release

PR #748, opened 2023-12-18
rely on dotnet cli to execute tool

### LINKED ISSUE #722: consider use ${userHome} instead while init
https://code.visualstudio.com/docs/editor/variables-reference#_predefined-variables
${userHome} - the path of the user's home folder

image

carlin-q-scott:
Fixes #722

### DIFF
--- a/src/Dotnet.Script.Core/Templates/launch.json.template
+++ b/src/Dotnet.Script.Core/Templates/launch.json.template
@@ -1,4 +1,4 @@
-﻿{
+{
     "version": "0.2.0",
     "configurations": [
         {
@@ -7,9 +7,8 @@
             "request": "launch",
             "program": "dotnet",
             "args": [
-                "exec",
-                "PATH_TO_DOTNET-SCRIPT",
-                "${file}"
+                "script",
+                "${relativeFile}"
             ],
             "cwd": "${workspaceRoot}",
             "stopAtEntry": false


filipw:
I don't think this is a good change - when it's installed as a global dotnet tool, then this template is used https://github.com/dotnet-script/dotnet-script/blob/master/src/Dotnet.Script.Core/Templates/globaltool.launch.json.template

I think the bug is rather that we check if the tool is globally installed by comparing if its path contains `dotnet/.tools` which will not match on Windows due to the path separator being wrong 😄

carlin-q-scott:
@filipw I have the tool globally installed and it used the local launch.json.template to init my project. Does it select the template based on whether the command is invoked via `dotnet-script init`, or `dotnet script init`?

carlin-q-scott:
@filipw Sorry about my earlier comment. I didn't understand that you were saying the template selection is based on the existence of a ~/dotnet/.tools directory. Windows supports both path separators as of Windows 10, but your period is in the wrong place. Global tools are installed at `%USERPROFILE%/.dotnet/tools` on Windows.

filipw:
Here is the code https://github.com/dotnet-script/dotnet-script/blob/be499540b8895dbc2902771e3fdb259d97235adc/src/Dotnet.Script.Core/Scaffolder.cs#L144-L145

we literally check for `.dotnet/tools` string when comparing the path, which will not match on Windows irrespective of whether it supports both path separators, as this is just the string comparison. I would say this is where it should be fixed.

carlin-q-scott:
I see. Yeah that won't work on Windows.

PR #753, opened 2024-05-06
fixed global tool path detection

filipw:
Fixes https://github.com/dotnet-script/dotnet-script/issues/697#issuecomment-2094336137

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -23,7 +23,7 @@ jobs:
         run: dotnet-script build/Build.csx
 
   build-mac:
-    runs-on: macos-latest
+    runs-on: macos-13
 
     steps:
       - uses: actions/checkout@v3
--- a/src/Dotnet.Script.Core/Scaffolder.cs
+++ b/src/Dotnet.Script.Core/Scaffolder.cs
@@ -141,7 +141,7 @@ namespace Dotnet.Script.Core
             _scriptConsole.WriteNormal("Creating VS Code launch configuration file");
             string pathToLaunchFile = Path.Combine(vsCodeDirectory, "launch.json");
             string installLocation = _scriptEnvironment.InstallLocation;
-            bool isInstalledAsGlobalTool = installLocation.Contains(".dotnet/tools", StringComparison.OrdinalIgnoreCase);
+            bool isInstalledAsGlobalTool = installLocation.Contains($".dotnet{Path.DirectorySeparatorChar}tools", StringComparison.OrdinalIgnoreCase);
             string dotnetScriptPath = Path.Combine(installLocation, "dotnet-script.dll").Replace(@"\", "/");
             string launchFileContent;
             if (!File.Exists(pathToLaunchFile))
--- a/src/Dotnet.Script.Tests/ScaffoldingTests.cs
+++ b/src/Dotnet.Script.Tests/ScaffoldingTests.cs
@@ -179,7 +179,7 @@ namespace Dotnet.Script.Tests
         [Fact]
         public void ShouldCreateUnifiedLaunchFileWhenInstalledAsGlobalTool()
         {
-            Scaffolder scaffolder = CreateTestScaffolder("somefolder/.dotnet/tools/dotnet-script");
+            Scaffolder scaffolder = CreateTestScaffolder($"somefolder{Path.DirectorySeparatorChar}.dotnet{Path.DirectorySeparatorChar}tools{Path.DirectorySeparatorChar}dotnet-script");
 
             using var scriptFolder = new DisposableFolder();
             scaffolder.InitializerFolder("main.csx", scriptFolder.Path);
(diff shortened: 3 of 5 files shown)


filipw:
The Mac failure is unrelated to this PR, I wonder what is going on

src/Dotnet.Script.Tests/ScriptExecutionTests.cs
```diff
@@ -259,14 +261,6 @@ public void ShouldExecuteRemoteScript(string url, string output)
             Assert.Contains(output, result.Output);
         }
 
-        [Fact]
-        public void ShouldExecuteRemoteScriptUsingTinyUrl()
```
filipw, line 263: tinyurl implemented captcha using cloudflare and we can't use it anymore 😅

however I am not sure why this test even exists, I assume to test if would follow redirects? if that is the case, then I added an `http://` variant to the test above, since github returns 301 then

.github/workflows/main.yml
```diff
@@ -23,7 +23,7 @@ jobs:
         run: dotnet-script build/Build.csx
 
   build-mac:
-    runs-on: macos-latest
+    runs-on: macos-13
```
filipw, line 26: this forces us back into intel Macs for now

seesharper (approved):
Looks good now 👍. Thanks

src/Dotnet.Script.Core/Scaffolder.cs
```diff
@@ -141,7 +141,7 @@ private void CreateLaunchConfiguration(string currentWorkingDirectory)
             _scriptConsole.WriteNormal("Creating VS Code launch configuration file");
             string pathToLaunchFile = Path.Combine(vsCodeDirectory, "launch.json");
             string installLocation = _scriptEnvironment.InstallLocation;
-            bool isInstalledAsGlobalTool = installLocation.Contains(".dotnet/tools", StringComparison.OrdinalIgnoreCase);
+            bool isInstalledAsGlobalTool = installLocation.Contains($".dotnet{Path.DirectorySeparatorChar}tools", StringComparison.OrdinalIgnoreCase);
```
IanKemp, line 144: Better to use `Path.Combine(".dotnet", "tools")`.

src/Dotnet.Script.Tests/ScaffoldingTests.cs
```diff
@@ -179,7 +179,7 @@ public void ShouldUpdatePathToDotnetScript()
         [Fact]
         public void ShouldCreateUnifiedLaunchFileWhenInstalledAsGlobalTool()
         {
-            Scaffolder scaffolder = CreateTestScaffolder("somefolder/.dotnet/tools/dotnet-script");
+            Scaffolder scaffolder = CreateTestScaffolder($"somefolder{Path.DirectorySeparatorChar}.dotnet{Path.DirectorySeparatorChar}tools{Path.DirectorySeparatorChar}dotnet-script");
```
IanKemp, line 182: Again, `Path.Combine` is your friend.

src/Dotnet.Script.Tests/ScriptExecutionTests.cs
```diff
@@ -250,7 +250,9 @@ public void ShouldSupportInlineNugetReferencesWithTrailingSemicoloninEvaluatedCo
 
         [Theory]
         [InlineData("https://gist.githubusercontent.com/seesharper/5d6859509ea8364a1fdf66bbf5b7923d/raw/0a32bac2c3ea807f9379a38e251d93e39c8131cb/HelloWorld.csx",
-                    "Hello World")]
+            "Hello World")]
+        [InlineData("http://gist.githubusercontent.com/seesharper/5d6859509ea8364a1fdf66bbf5b7923d/raw/0a32bac2c3ea807f9379a38e251d93e39c8131cb/HelloWorld.csx",
```
IanKemp, line 254: Should be `https://`

src/Dotnet.Script.Core/Scaffolder.cs
```diff
@@ -141,7 +141,7 @@ private void CreateLaunchConfiguration(string currentWorkingDirectory)
...
             string dotnetScriptPath = Path.Combine(installLocation, "dotnet-script.dll").Replace(@"\", "/");
```
IanKemp, line 145: @seesharper why does this normalise everything to forward-slash, instead of using the platform-specific separator?

filipw: it doesn't - `DirectorySeparatorChar` resolves to the platform specific separator

src/Dotnet.Script.Core/Scaffolder.cs
filipw: internally it uses `DirectorySeparatorChar` too so it's a matter of preference

src/Dotnet.Script.Tests/ScriptExecutionTests.cs
filipw: This is intentionally http because this variant tests following a redirect (which would be issued from http to https)

PR #770, opened 2024-12-06
upgrade roslyn packages

WeihanLi:
upgrade roslyn packages


filipw:
Thanks, but the packages are intentionally not revved up to 4.12.0 yet, because they are aligned with OmniSharp, which is [not on 4.12.0 yet](https://github.com/OmniSharp/omnisharp-roslyn/blob/master/Directory.Packages.props#L12)

PR #596, opened 2020-11-23
cache script per file instead of per directory

DanielEgbers:
## Description

Currently only one script per directory is cached.

This PR adds the script file name into the cache directory path, therefore caching scripts per file instead of per directory.

Example:
|Script|Current Cache|New Cache|
|-|-|-|
|```D:\scriptA.csx```|```%tmp%\dotnet-script\D\execution-cache\script.dll```|```%tmp%\dotnet-script\D\scriptA.csx\execution-cache\script.dll```|
|```D:\scriptB.csx```|```%tmp%\dotnet-script\D\execution-cache\script.dll```|```%tmp%\dotnet-script\D\scriptB.csx\execution-cache\script.dll```|

## Motivation and Context
I would like to benefit from the execution speed up by caching without having to separate every script into its own directory.

Issue #540, opened 2020-06-01
Execution cache file is per script folder leading to error when running scripts in parallel


mungojam:
I'm using dotnet script for some scripts which I run in parallel. Both scripts are in the same folder, but unless I turn off caching, this leads to a clash on the common cache file `script.dll`:

> The process cannot access the file 'c:\somewhere\scripts\execution-cache\script.dll' because it is being used by another process.

The first script to get there works fine, but the 2nd fails because it tries to use the cache at the same time.

~It works if I use `--no-cache` but I would rather not.~

mungojam:
Oddly, it doesn't even work when I use `--no-cache`:

> dotnet script --no-cache .\somewhere\configure-backend.csx 
> The process cannot access the file 'C:\Users\jenkins\AppData\Local\Temp\dotnet-script\C\somewhere\build-scripts\execution-cache\script.dll' because it is being used by another process.

seesharper:
Hi @mungojam 

The reason it probably won't work even with the `--no-cache` option is that we always run the script from a compiled script (script.dll).

That being said, the cache should definitely handle executing two scripts in parallell from the same folder. Looking into it👍

mungojam:
I've just spotted https://github.com/filipw/dotnet-script/issues/507 which is related.

I'm planning on submitting a PR for this one first though, using the filename in the temporary folder path if that makes sense?

kuacci:
Is it possible to implement an abstraction layer of file system just like what scriptcs did?

mganss:
A workaround is to set `DOTNET_SCRIPT_CACHE_LOCATION` to a separate directory for each of the scripts you are running in parallel.

mungojam:
Aha! Neat idea, thanks. I can then put the cache within the build workspace
too so it's easier to find and archive as needed for build diagnostics

mungojam:
Fixed by #596

Issue #507, opened 2020-01-04
Cannot run multiple scripts (which are at the same location) in parallel


ejerome:
Hello,

Thanks for your cool project which is very useful !
It seems that, we cannot have a set of scripts that could be instantiated multiple times & run concurrently from the same location.
![dotnet-script_issueMultipleScriptSameLocation](https://user-images.githubusercontent.com/5173267/71756222-ed518d80-2e8e-11ea-9273-7762ee1f8886.JPG)

It seem that [dotnet-script/src/Dotnet.Script.Core/Commands/ExecuteScriptCommand.cs](https://github.com/filipw/dotnet-script/blob/0848b0469146293b86a35798c8fe7711b0f1b91a/src/Dotnet.Script.Core/Commands/ExecuteScriptCommand.cs) `GetLibrary` method have an hardcoded 'default single name' for all scripts of the same location path which is : `script.dll`.

Is there a specific reason for that or are you able to use the scriptFileName or a GUID as unique identifier for example for your next release ?

Thanks a lot !

DanielEgbers:
This is now fixed in v1.1.0 by #596. You can now run multiple scripts from the same folder in parallel.

Only if you run both scripts not cached and at the very same time, it could break because the they still share the same folder for building the assembly. This can not be changed, because it is required for OmniSharp to work.

PR #778, opened 2025-10-30
Added '--cache-path' command line option

psryland:
Added a new command-line option "--cache-path" (or "-p") to allow users to specify the temporary directory used to build the script. This is equivalent to the 'DOTNET_SCRIPT_CACHE_LOCATION' environment variable but is far more convenient when scripts are run as part of MSBuild.  In particular, parallel runs of a script can specify different cache paths preventing the "script.csproj is in use" error that can happen. (Re: issues #507, #596, #540)

For example, 
```xml
    <Exec
      Command="dotnet-script --cache-path $(OutDir) PostBuildStep.csx -- $(TargetPath)"
      WorkingDirectory="$(ProjectDir)"
    />
```

### DIFF
--- a/src/Dotnet.Script.Tests/ExecutionCacheTests.cs
+++ b/src/Dotnet.Script.Tests/ExecutionCacheTests.cs
@@ -151,7 +151,7 @@ namespace Dotnet.Script.Tests
 
         private static string GetPathToExecutionCache(string pathToScript)
         {
-            var pathToTempFolder = Dotnet.Script.DependencyModel.ProjectSystem.FileUtils.GetPathToScriptTempFolder(pathToScript);
+            var pathToTempFolder = Dotnet.Script.DependencyModel.ProjectSystem.FileUtils.GetPathToScriptTempFolder(pathToScript, cachePath: null);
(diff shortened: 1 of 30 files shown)


src/Dotnet.Script.Tests/ExecutionCacheTests.cs
```diff
@@ -133,6 +133,34 @@ public void ShouldCacheScriptsFromSameFolderIndividually()
             Assert.False(thirdResultOfScriptB.Cached);
         }
 
+        [Fact]
```
filipw: could you additionally update these tests so that all of them are executed for both default cache path and custom path?

filipw:
Thanks a lot for the PR! I think this is a very welcome addition 😄

filipw:
The linux test failures are unrelated to your PR. They will be resolved once we merge #779

src/Dotnet.Script.Tests/ExecutionCacheTests.cs
psryland: Yep, will do.

